PR #57, opened 2026-03-01
feat: add line_wrap config option for preview pane

### LINKED ISSUE #47: Feature: optional line wrapping in preview pane
## Summary

Long lines in the preview pane are truncated with `…`. It would be great to have an optional line-wrap mode so the full output is visible without needing a very wide terminal.

## Current behavior

`pane_preview.rs` calls `truncate_line(line, available_width)` which cuts lines at the pane border and appends `…`. This is clean for short outputs but makes it hard to read tool calls, file paths, and long prose that Claude Code often produces.

## Proposed behavior

A config toggle like:

```toml
[ui]
wrap_preview = true   # default: false (current truncation behavior)
```

When enabled, use ratatui's `Paragraph::wrap(Wrap { trim: false })` instead of manual truncation. Scrolling would need to account for wrapped line heights.

## Context

Running 7 Claude Code agents in detached tmux sessions, using tmai as a monitoring dashboard. The preview is the primary way to check agent state without attaching, so seeing full output matters.

## Alternatives considered

- Wider terminal — helps but limited by screen
- `p` passthrough mode — works for interaction but loses the dashboard overview
- Attaching directly — defeats the purpose of tmai as a dashboard

gcamilo:
## Summary

Adds a `line_wrap` boolean option under `[ui]` in the config file (default: `false`). When enabled, the preview pane wraps long lines using ratatui's `Paragraph::wrap(Wrap { trim: false })` instead of truncating them with an ellipsis character.

This makes it much easier to read agent output that contains long lines (e.g., Claude Code tool calls, JSON payloads, log messages) without losing content off-screen.

### Config

```toml
[ui]
line_wrap = true
```

## Changes

- **`crates/tmai-core/src/config/settings.rs`** — Add `line_wrap: bool` to `UiSettings` (serde default `false`)
- **`crates/tmai-core/src/state/store.rs`** — Add `line_wrap: bool` to `AppState`
- **`src/ui/app.rs`** — Wire setting from config into app state
- **`src/ui/components/pane_preview.rs`** — Skip `truncate_line()` when wrapping; apply `Paragraph::wrap(Wrap { trim: false })` conditionally

## Test plan

- [x] All 80 existing tests pass
- [x] Manual testing: long lines wrap correctly in preview pane
- [x] Manual testing: `line_wrap = false` (default) preserves existing truncation behavior
- [ ] Verify no visual regressions with ANSI-colored content when wrapping

Closes #47

## Summary by CodeRabbit

* **新機能**
  * プレビューペインにテキスト自動折り返し機能を追加。長い行が複数行に折り返されスクロール表示されます。
  * 折り返しのオン／オフはUI設定から切替可能で、初期設定は無効（オフ）です。


trust-delta (commented):
Thanks for the contribution, @gcamilo! This is the first external PR for tmai — really appreciate you taking the time to implement this. 🎉

The `line_wrap` feature itself looks solid: clean config integration, sensible default (`false`), and the visual-line scroll estimation works well.

A few points:

### 1. `layout.rs` change should be a separate PR (blocking)

The `with_preview_height()` fix is unrelated to `line_wrap` — it changes `preview_height` from a no-op to actually taking effect. This has a visible side effect for existing users:

- Default `preview_height = 50` was previously ignored (hardcoded to `session_list_width_pct = 35`)
- With this change, the default layout shifts from 35/65 to 50/50
- `session_list_height_pct` also changes from 25 → 50 (affects vertical split)

Could you split this into a separate PR? It's a valid fix, but the blast radius is different from `line_wrap`.

### 2. Border color duplication (nitpick)

The `AgentStatus → Color` mapping is now in two places (wrap path L119-125 and non-wrap path L169-176). A small helper like `fn border_color(status: &AgentStatus) -> Color` would keep them in sync. Happy to fix this on our side if you prefer.

### 3. Per-frame full content join (minor, no action needed now)

When wrapping, `content_lines[..total_lines].join("\n")` + `into_text()` runs every frame on the full content. With typical capture-pane sizes (~300 lines) this is fine, but if we ever increase the buffer, a tail-only approach would be better. Not blocking for this PR.

---

With the `layout.rs` change split out, this is good to merge. Thanks again!

trust-delta:
Thanks for the updates, @gcamilo! The layout.rs fix is already merged via #64.

This looks good — clean separation of wrap/non-wrap paths, `border_color` extracted, and the visual line calculation using unicode_width on styled text is the right approach.

Merging!

PR #373, opened 2026-04-12
feat: PrCiPassed auto-dispatch_review and tri-state notification UI (#364 part 3/3)

trust-delta:
Final PR of #364 (part 3/3). PR-A (#367) introduced `EventHandling`; PR-B (#371) added the `AutoActionExecutor` service with `CiFailed` and `PrReviewFeedback` handlers. This PR wires the `PrCiPassed` handler and completes the WebUI tri-state notification UI.

## Summary

- **PrCiPassed auto-dispatch_review**: when `on_ci_passed == AutoAction`, tmai dispatches a reviewer iff the PR has no active reviewer recorded in `.task-meta/{branch}.json` **and** `get_pr_comments` returns no review comments. Respects the existing `max_review_loops` guardrail and escalates via `GuardrailExceeded` on overflow. Reuses the refactored `perform_dispatch_review` via an injected `ReviewDispatcher` trait so there is a single code path shared with `/api/review/dispatch`.
- **Task-meta lookup by PR**: added `task_meta::store::find_by_pr` used by the PrCiPassed handler (PR-B only had lookup by branch).
- **Settings serde tightening**: `EventHandling` now serializes as `"off" | "notify" | "auto_action"` (matches the WebUI TypeScript union). Deserialization still accepts the legacy `"notify_orchestrator"` spelling to keep existing `config.toml` readable.
- **Settings API**: `NotifySettingsResponse` / `UpdateNotifySettingsRequest` switched from `bool` to `EventHandling`. GET and PUT now include `auto_action_templates` (partial updates supported).
- **WebUI**: the per-event notify toggle is now a tri-state button group. Events that have an AutoAction handler (`on_ci_failed`, `on_pr_comment`, `on_ci_passed`) render three options; all other events render two (`Off` / `Notify`). Conditional rendering was chosen over a disabled third option for cleaner visual scanning. AutoAction mode reveals an inline template editor for the two events with per-role prompts; PrCiPassed has none because `dispatch_review` composes its own reviewer prompt.

## Test plan
- [x] `cargo test -p tmai-core auto_action` — 34 tests pass (includes 5 new PrCiPassed cases + `find_by_pr`).
- [x] `cargo clippy --all-targets -- -D warnings` clean.
- [x] `cargo fmt` clean.
- [x] `npm run build` in `crates/tmai-app/web` — bundle regenerated and committed.
- [x] Biome check clean (pre-existing `biome.json` schema info only).
[...]

PR #374, opened 2026-04-12
feat: buffer orchestrator notifications during busy state (#372)

### LINKED ISSUE #372: feat: buffer orchestrator notifications during busy state and flush on idle
## Problem

\`OrchestratorNotifier\` filters events to **idle orchestrator agents only**. When the orchestrator is busy (working on a task), any notifications that would have been sent are **dropped entirely**, not queued.

Log evidence (captured 2026-04-12):

\`\`\`
DEBUG No idle orchestrator agents to notify [tmai] Agent \"37\" has stopped.
\`\`\`

Real impact during dogfooding: while the orchestrator was investigating a different issue, a worker completed its PR and stopped. The \`AgentStopped\` notification was dropped, so the orchestrator did not learn about the completion until it manually polled (via \`/loop\`). In isolation this is tolerable, but over a long session the silent loss of important events (CI failures, PR comments, agent stops, guardrail breaches) degrades situational awareness.

## Proposed design

Buffer notifications per-orchestrator while busy. On transition to idle, flush the buffered notifications as a single combined message.

### Data

\`\`\`rust
// SharedState
pub orchestrator_notify_buffer: HashMap<String, Vec<BufferedNotification>>,

struct BufferedNotification {
    message: String,
    source: String,        // e.g. \"action-dispatch_issue\", for dedup
    timestamp: Instant,    // for TTL
}
\`\`\`

### Flow

1. \`OrchestratorNotifier\` processes event as today (resolves target orchestrator, builds message)
2. If target orchestrator is **idle** → emit \`PromptReady\` immediately (current behavior)
3. If target is **busy** → append to \`orchestrator_notify_buffer[target]\` instead of dropping
4. Subscribe to \`CoreEvent::AgentStatusChanged\` — when an orchestrator transitions to \`Idle\`:
   - Take any buffered messages for that orchestrator
   - Combine with \`\\n\\n---\\n\\n\` separator
   - Emit a single \`PromptReady\` with the combined message

### Guardrails

| Concern | Mitigation |
|---|---|
| Unbounded memory | Cap buffer at N messages or M bytes per orchestrator (suggest 20 / 10 KB). Oldest dropped on overflow |
| Stale notifications | TTL (suggest 10 min). Expired entries dropped when buffer is touched or on next flush |
| Duplicate events | Dedup by \`source\` string — if same key already present, replace (latest wins) or drop (first wins). Decide based on event kind |
| Persistence | Memory-only is fine (tmai restart = fresh start; acceptable) |
| Flush on session start | When an orchestrator is newly detected (\`SessionStart\` hook) and has buffered messages from a previous lifecycle, flush them |

### Separation from \`prompt_queue\`

\`AppState::prompt_queue: HashMap<String, VecDeque<String>>\` already exists for \`send_prompt\` queuing (active instructions when agent is busy). Do **not** merge this feature into that queue — semantics differ:

- \`prompt_queue\` → active instructions, one-at-a-time delivery
- \`orchestrator_notify_buffer\` → passive notifications, batched delivery

Keeping them separate allows independent size caps, TTLs, and flush triggers.

## Implementation sketch

- **Module:** extend \`crates/tmai-core/src/orchestrator_notify/\`
- **Service:** same subscriber, add buffer logic + a second event handler for \`AgentStatusChanged\`
- **Config:** add \`OrchestratorNotifySettings::buffer_when_busy: bool\` (default \`true\`), \`buffer_ttl_secs: u64\` (default 600), \`buffer_max_messages: usize\` (default 20)
- **Tests:** unit tests for buffer append / overflow / TTL expiry / dedup / flush on idle

## Related

- Partially overlaps with \#266 (bidirectional communication) but the solution space is different — this is about **not losing** outbound notifications, not about **new inbound channels**.
- Complementary to \#364 (AutoActionExecutor) — AutoActionExecutor reduces the number of events that need orchestrator attention; buffering ensures the remaining ones aren't lost.

trust-delta:
## Summary

Resolves #372. When all orchestrators are busy, outbound notifications from `OrchestratorNotifier` were being dropped with a `debug!("No idle orchestrator agents to notify ...")` log. This PR buffers them per-orchestrator and flushes on transition to idle (or on orchestrator re-appearance), so important events (`AgentStopped`, `PrCiFailed`, `GuardrailExceeded`, etc.) are no longer silently lost.

## Behaviour

- **Append path** — on each event, orchestrators are split into idle (send `PromptReady` now) and busy (append to buffer keyed by orchestrator target). If `buffer_when_busy = false`, the legacy drop-with-log behaviour is preserved.
- **Dedup** — within one orchestrator's buffer, a new entry whose `source` matches an existing entry replaces it (latest wins). So repeated "Agent X stopped" from the same source collapses.
- **TTL** — entries older than `buffer_ttl_secs` (default 600 s) are dropped on next append or flush.
- **Overflow** — when a buffer reaches `buffer_max_messages` (default 20), the oldest entry is dropped before append.
- **Flush triggers** — `CoreEvent::AgentStatusChanged { new_status: "idle" }` AND `CoreEvent::AgentAppeared` for an orchestrator cause its buffer to be taken, TTL-swept, sorted by timestamp, joined with `\n\n---\n\n`, and emitted as a single `PromptReady`.

## Target selection while busy

When a notification has no idle orchestrator to send to, the buffer is keyed by the **same orchestrator target the existing loop already iterates over** — i.e. every non-self, non-virtual orchestrator agent (now regardless of idle state). This matches the previous broadcast-to-all-orchestrators intent and keeps the notifier's target resolution single-sourced. The alternative ("buffer per project path, flush to whichever idle comes first") was not adopted — it would have required a second target-resolution path.

## Shared-Arc confirmation (review reminder)

I confirmed the new `orchestrator_notify_buffer` field is shared via `SharedState` between the write path (notifier busy branch) and the read/flush path (flush-on-idle / flush-on-appear), using clones of the same `Arc`, not separate `Arc::new` calls per site. `new_shared_buffer()` is called exactly once per spawn site in `src/main.rs` (TUI + wrap/web paths), stored in `AppState::orchestrator_notify_buffer`, and the same handle is passed to `OrchestratorNotifier::spawn`. The service holds it for its entire lifetime; both the append branch and the flush branch use that single handle.

## Hot-reload

`buffer_when_busy` / `buffer_ttl_secs` / `buffer_max_messages` live inside `OrchestratorNotifySettings`, so they are already covered by the existing `*ns.write() = updated.notify` block in `update_orchestrator_settings` (src/web/api.rs). I added a comment in the settings-merge code noting this and made the merge inherit current values (no REST field added yet — a follow-up PR can expose these in the WebUI if desired).

## Tests

Added under `orchestrator_notify::{buffer,service}::tests`:

- Pure buffer: append accumulates, dedup by source, overflow drops oldest, TTL expiry sweep (using `Instant::now() - Duration` to inject expired entries — no clock injection needed), take removes key, combine separator, multi-target isolation
- Service-level: `test_busy_orchestrator_buffers_instead_of_dropping`, `test_flush_on_idle_transition_delivers_combined_prompt`, `test_flush_on_agent_appeared_for_orchestrator`, `test_buffer_when_busy_false_falls_back_to_drop`, `test_multiple_orchestrators_buffered_independently`

All 48 tests in the `orchestrator_notify` module pass. TTL was tested via pre-seeded expired `Instant`s, so no clock-injection plumbing was needed.

## Out of scope

- No changes to `AppState::prompt_queue` (separate semantics per issue design)
- No changes to `AutoActionExecutor` emit path
- No persistence (memory-only per decision — tmai restart = fresh start)
- No new UI for the three config fields (they are hot-reloadable through the existing settings API)

## Test plan

- [x] `cargo fmt` clean
- [x] `cargo clippy -- -D warnings` clean
- [x] `cargo test -p tmai-core --lib orchestrator_notify` — 48/48 pass
- [x] `cargo build` — workspace builds
- [ ] Manual dogfooding: let a worker stop while orchestrator is processing an unrelated task; confirm the stop notification arrives as a combined prompt when the orchestrator returns to idle

## Summary by CodeRabbit

* **New Features**
  * オーケストレーター向け通知バッファを追加。忙しい時の通知を蓄積し、結合して後で送信可能（有効化・TTL・最大件数設定あり）。
* **Behavior Changes**
  * アイドルへは即時通知、繁忙時は設定に応じてバッファまたは破棄。アクション通知は識別子で重複抑制。
* **API / UI**
  * 通知バッファ関連設定（有効化・TTL・最大件数）を取得・更新可能に。
* **Bug Fix / State**
  * PID再割当時にバッファ内のキーを移行して整合性を維持。


trust-delta (commented):
LGTM — all critical checks (A–H) verified.

**A. Shared handle ✅** — `SharedNotifyBuffer = Arc<RwLock<HashMap<..>>>` created once per process in `src/main.rs:222-223` (tmux) and `src/main.rs:460-461` (webui), stashed in `AppState.orchestrator_notify_buffer` (store.rs:554), and the same `Arc` is cloned into `OrchestratorNotifier::spawn`. Producer (busy branch, service.rs:~138) and consumer (flush, service.rs:~180) both read/write via the handle received as a parameter — no per-callsite `Arc::new`. Doesn't repeat the #373 AutoActionExecutor mistake.

**B. Overflow/dedup/TTL ✅** — FIFO drop-oldest via `entries.remove(0)` (buffer.rs:113-115), latest-wins dedup by `source` (buffer.rs:107-110), `std::time::Instant` (monotonic) everywhere, TTL sweep on both append and flush, `take_for_flush` returns empty vec (no spurious `PromptReady`) when all entries expired.

**C. Flush triggers ✅** — `AgentStatusChanged { new_status == "idle" }` and `AgentAppeared` in `maybe_flush_for_event`. `AgentAppeared` covers the tmai-restart case per the issue spec. Non-orchestrator / non-idle agents are filtered via a second `is_orch_idle` check so the two triggers don't fire redundantly on the same transition.

**D. Target identity ✅** — both append and flush key off the `agents` map key. Append iterates `st.agents` and uses the map key as target; flush resolves via `st.agents.get(target)` with the same key from the event payload. No pane_id/stable_id divergence.

**E. Concurrency ✅** — `parking_lot::RwLock`, no `.unwrap()` on locks, append + flush happen serially in the single `event_rx` loop, and no other codepath touches the buffer (verified via file list: only `buffer.rs`, `service.rs`, `store.rs`, `main.rs` reference it; web/api.rs only hot-reloads the settings struct).

**F. Config compat ✅** — all three new fields have `#[serde(default = ...)]`; existing TOML loads unchanged. Hot-reload piggy-backs on the existing `*ns.write() = updated.notify` path in `update_orchestrator_settings` (web/api.rs:1609-1614 preserve request-absent values from current settings). No extra wiring needed.

**G. Scope ✅** — touches only orchestrator_notify + its config + state handle + two main.rs spawn sites + one web/api.rs struct literal. No `prompt_queue`, no `AutoActionExecutor`, no disk persistence.

**H. Tests ✅** — buffer.rs unit tests cover append/dedup/overflow/TTL/take-removes-key/combine/multi-target. service.rs tests cover: busy-buffers (#1), flush-on-idle combined prompt (#5), flush-on-AgentAppeared (#6), `buffer_when_busy=false` drop (#7), multi-orchestrator isolation (#8). TTL-in-service (#3) is reasonably skipped given clock-injection cost, covered instead by buffer unit test `ttl_expires_on_append_sweep`.

CI: all 7 checks green. No blockers.

PR #375, opened 2026-04-12
feat: inject --permission-mode for dispatched workers (default acceptEdits)

trust-delta:
## Summary

Inject Claude Code's \`--permission-mode acceptEdits\` flag when spawning workers via \`dispatch_issue\` / \`dispatch_review\`. File edits flow without per-tool approval prompts; Bash continues through the existing tmai auto-approve pipeline (AI judge guardrail intact).

Motivation: recent Claude Code versions added \`acceptEdits\` as a permission level that auto-approves Edit/Write/MultiEdit tools while still prompting for Bash. For tmai-dispatched implementation workers, this is the sweet spot — it removes the friction of per-file-edit approvals without weakening shell-command gatekeeping.

## Scope

Applies to:
- \`spawn_worktree\` (the internal handler used by \`dispatch_issue\`)
- \`dispatch_review\`

Does **not** apply to:
- Manual \`spawn_agent\` calls
- \`spawn_orchestrator\` — the orchestrator inherits whatever mode the user picked when they launched Claude Code

## Design

- \`SpawnSettings.worker_permission_mode: String\`
  - Default: \`\"acceptEdits\"\`
  - Valid: \`\"default\" | \"plan\" | \"acceptEdits\" | \"dontAsk\"\`
  - Setting value \`\"default\"\` (or empty) skips injection entirely, leaving Claude Code's own default in effect
- \`GET /api/settings/spawn\` exposes \`worker_permission_mode\`
- \`PUT /api/settings/spawn\` accepts optional updates, validates against the known set server-side before persisting
- WebUI adds a dropdown under Settings → Spawn with a short safety note

## Rationale / alignment

- tmai stays **outside the AI** (per project design principle — see recent memory). This change rides the CLI flag; no SDK integration.
- Bash gatekeeping is unchanged. tmai auto-approve AI judge still runs on Bash invocations, and existing \`max_ci_retries\` / \`max_review_loops\` guardrails remain in force.
- Orchestrator is untouched because its behaviour should match the user's own Claude Code CLI configuration, not tmai's defaults.

## Test plan

- [x] \`cargo fmt && cargo clippy -- -D warnings\` clean
- [x] \`cargo test -p tmai-core --lib config\` passes
- [x] \`npm run build\` produces updated bundle (committed)
- [ ] Dogfooding: dispatch an issue; verify \`claude --permission-mode acceptEdits\` in the spawned pane and that file-edit tool calls no longer prompt, while Bash still does


trust-delta (commented):
LGTM — `--permission-mode acceptEdits` injection is cleanly scoped to dispatched workers.

**Verified against checklist**
- Injection scope: only `perform_dispatch_review` (src/web/api.rs:3079) and `spawn_worktree` (src/web/api.rs:3253). `spawn_orchestrator` (src/web/api.rs:2339) and low-level `spawn_agent` / `spawn_in_tmux` / `spawn_in_pty` paths are untouched. ✅
- Mode handling: `"default"` / empty → flag omitted; `"plan" | "acceptEdits" | "dontAsk"` → `--permission-mode <v>` prepended before the positional prompt (src/web/api.rs:3254-3258, 3080-3083). ✅
- PUT validation: values outside `["default","plan","acceptEdits","dontAsk"]` are not persisted (src/web/api.rs:1242-1253). ✅
- Defaults agree: `default_worker_permission_mode()` and `impl Default for SpawnSettings` both yield `"acceptEdits"` (crates/tmai-core/src/config/settings.rs:826-851). ✅
- Backward compat: `#[serde(default = "default_worker_permission_mode")]` on the field — existing configs load cleanly. ✅
- WebUI: dropdown shows all 4 options with `acceptEdits (recommended)`, persists on change (no Save button); `WorkerPermissionMode` exported from `api-http.ts` and re-exported via `api-tauri.ts`. ✅
- Bundle freshness: `index.html` points at `index-DEZqvxU1.js`; old `index-B0V4RXHx.js` replaced. ✅

**Scope boundaries respected**
No SDK/Skill integration, no AutoApprove pipeline changes, orchestrator spawn unchanged, no new permission-mode values beyond CC's accepted set.

**Minor observations (non-blocking)**
- `update_spawn_settings` (src/web/api.rs:1242-1253) silently ignores invalid `worker_permission_mode` values rather than returning `400 Bad Request`. The TS client constrains the union type, so UI cannot send garbage, but direct HTTP callers get a 200 with the value effectively dropped. Consider surfacing a `400` on unknown values in a follow-up.
- **Test gap**: no unit test added (e.g. `SpawnSettings::default().worker_permission_mode == "acceptEdits"`, or arg-vector construction in `spawn_worktree`/`perform_dispatch_review`). Worth adding a small config test; full dispatch-path test is reasonably out of scope.

Not merging per instructions.

PR #379, opened 2026-04-12
fix: save full OrchestratorSettings to config.toml — notify/guardrails/pr_monitor/auto_action_templates persist (#376)

### LINKED ISSUE #376: fix: save_project_orchestrator drops notify/guardrails/pr_monitor settings on global save
## Problem

\`Settings::save_project_orchestrator\` (\`crates/tmai-core/src/config/settings.rs:1523\`) only writes \`enabled\`, \`role\`, and \`rules.*\` to the global \`[orchestrator]\` section of \`config.toml\`.

It silently **drops** on save:

- \`notify.*\` (\`on_ci_failed\`, \`on_ci_passed\`, \`on_pr_comment\`, buffer settings, templates, …)
- \`guardrails.*\` (\`max_ci_retries\`, \`max_review_loops\`, \`escalate_to_human_after\`)
- \`auto_action_templates.*\`
- \`pr_monitor_enabled\`
- \`pr_monitor_interval_secs\`

\`update_orchestrator_settings\` applies the values to in-memory \`Arc<RwLock<_>>\` handles (hot-reload), then calls \`reload_settings()\` which re-reads the disk — at which point the un-persisted fields snap back to their defaults. Users see the WebUI value "save" and work for the current session, then revert on next tmai startup.

\`pr_monitor_enabled\` is the most visible symptom because:
1. PR Monitor service only starts if \`pr_monitor_enabled == true\` at tmai launch time
2. Toggling it via WebUI does not persist
3. So PR/CI notifications never work after tmai restart even if user enabled them in the UI

## Reproduction

\`\`\`
# with orchestrator.pr_monitor_enabled unset in config.toml
$ curl -X PUT -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \\
       -d '{"pr_monitor_enabled": true}' \\
       http://localhost:9876/api/settings/orchestrator
{"ok": true}

$ curl -H "Authorization: Bearer $TOKEN" \\
       http://localhost:9876/api/settings/orchestrator | jq '.pr_monitor_enabled'
false  # silently reverted
\`\`\`

Same pattern for any \`notify.on_X\`, \`guardrails.max_X\`, or \`auto_action_templates\` fields when no per-project override is active.

## Root cause

The \`None\` branch of \`save_project_orchestrator\` (lines 1525-1561) hand-writes four \`save_toml_value\` calls (\`enabled\`, \`role\`, \`rules.branch\`, \`rules.merge\`, \`rules.review\`, \`rules.custom\`) and stops there. The \`Some(proj_path)\` branch (lines 1562-1581) rewrites the entire \`[[project]]\` entry so per-project saves work correctly.

## Fix

Either:

1. **Enumerate all fields** — add \`save_toml_value\` / \`save_toml_nested_value\` calls for every \`OrchestratorSettings\` field in the \`None\` branch. Scales poorly as the struct grows.
2. **Rewrite the whole \`[orchestrator]\` table** — serialize \`orch: &OrchestratorSettings\` to a \`toml_edit::Table\` via serde and splice it in under the \`[orchestrator]\` key. Keeps behaviour aligned with the per-project branch and auto-covers future field additions.

Option 2 is preferred. It mirrors the per-project save path which uses \`save_project_configs\` to rewrite the whole \`[[project]]\` array.

## Workaround (until fix lands)

Edit \`~/.config/tmai/config.toml\` manually and add missing fields under \`[orchestrator]\`, \`[orchestrator.notify]\`, \`[orchestrator.guardrails]\`, then restart tmai. WebUI edits stick for the current session but do not persist.

## Scope notes

This bug has been silently degrading the dogfooding experience — notifications that "worked" after a WebUI toggle stopped working after every tmai restart, making this issue hard to reproduce without log inspection.

Priority: **high** (user-facing data loss on save).

trust-delta:
## Summary

Fixes #376. `Settings::save_project_orchestrator(None, ...)` silently dropped every `OrchestratorSettings` field except `enabled`, `role`, and `rules.*` when writing the global `[orchestrator]` table. WebUI edits to `notify.*`, `guardrails.*`, `auto_action_templates.*`, `pr_monitor_enabled`, and `pr_monitor_interval_secs` took effect for the current session (via the in-memory `Arc<RwLock<_>>` hot-reload) but reverted on the next `reload_settings()` or tmai restart. `pr_monitor_enabled` was the most visible symptom — users toggled it on in the UI and PR/CI notifications silently stopped working after every restart.

## Approach

Rewrote the `None` branch to serialize the full `OrchestratorSettings` via a `{ orchestrator = <orch> }` wrapper and splice the resulting `Item::Table` into the existing `toml_edit::DocumentMut` under the `orchestrator` key. This mirrors how the per-project `Some(proj_path)` branch already works (via `save_project_configs`) and auto-covers future field additions — no hand-written `save_toml_value` / `save_toml_nested_value` chain to maintain.

Rejected the alternative of enumerating every field explicitly: brittle, scales poorly, and is exactly how this bug got shipped in the first place.

Assigning `doc["orchestrator"] = new_item` in `toml_edit` replaces the entire subtree (scalars + nested subtables), so stale entries from prior saves cannot leak through.

## Formatting / comment preservation

Unrelated sections (`[ui]`, top-level scalars, comments outside `[orchestrator]`) are preserved because we operate on a `DocumentMut` and only replace the `orchestrator` key. Comments **inside** the previous `[orchestrator]` block are lost on overwrite — this matches the existing behaviour of the per-project branch and is unavoidable when rewriting the whole table.

## New tests (`config::settings::tests`)

- `save_project_orchestrator_global_persists_all_fields` — save a fully non-default `OrchestratorSettings` (notify states, buffer settings, custom templates, guardrails, `auto_action_templates`, `pr_monitor_*`), reload via `toml::from_str`, assert `reloaded.orchestrator == orch`.
- `save_project_orchestrator_global_overwrites_previous_save` — save twice with different values; verify the second save does not leak stale custom templates from the first.
- `save_project_orchestrator_global_preserves_unrelated_sections` — seed `config.toml` with `poll_interval_ms` and `[ui]`, save orchestrator, verify unrelated values survive.
- `save_project_orchestrator_per_project_still_works` — regression guard for the `Some(proj_path)` branch: save into an existing `[[project]]` entry, verify the orchestrator override is written and other project entries are preserved.

All four tests use a `temp_env`-based helper that redirects `XDG_CONFIG_HOME` and `HOME` to an isolated tempdir, so they never touch the real user config.

## Scope

- Only the `None` branch changed. The `Some(proj_path)` branch is untouched.
- `OrchestratorSettings` struct, `save_toml_value` / `save_toml_nested_value` helpers, and the REST PUT handler in `src/web/api.rs` were all deliberately left alone.

## Test plan

- [x] `cargo fmt`
- [x] `cargo clippy --all-targets -- -D warnings`
- [x] `cargo test -p tmai-core --lib config::settings::` — 38 passed (including 4 new)
- [x] `cargo test -p tmai-core` — all settings tests pass; the single unrelated pre-existing failure in `detectors::claude_code::tests::test_multi_select_with_trailing_empty_lines` reproduces on `main` before this patch
- [ ] Manual WebUI: toggle `pr_monitor_enabled` on, restart tmai, confirm it stays on

## リリースノート

* **バグ修正**
  * オーケストレーター設定の保存処理を改善し、グローバル/プロジェクト単位の保存で設定の全項目が保持され、古いサブフィールドやテンプレートが残らないようにしました。既存の他セクションやコメントも維持されます。

* **テスト**
  * 設定の永続化と読み込みの回帰テストを追加し、保存→再読込での完全な等価性と上書き挙動を検証するようにしました。

### DIFF
--- a/crates/tmai-core/src/config/settings.rs
+++ b/crates/tmai-core/src/config/settings.rs
@@ -1569,41 +1569,64 @@ impl Settings {
     pub fn save_project_orchestrator(project_path: Option<&str>, orch: &OrchestratorSettings) {
         match project_path {
             None => {
-                // Save to global [orchestrator] section
-                Self::save_toml_value(
-                    "orchestrator",
-                    "enabled",
-                    toml_edit::Value::from(orch.enabled),
-                );
-                Self::save_toml_value(
-                    "orchestrator",
-                    "role",
-                    toml_edit::Value::from(orch.role.as_str()),
-                );
-                Self::save_toml_nested_value(
-                    "orchestrator",
-                    "rules",
-                    "branch",
-                    toml_edit::Value::from(orch.rules.branch.as_str()),
-                );
-                Self::save_toml_nested_value(
-                    "orchestrator",
-                    "rules",
-                    "merge",
-                    toml_edit::Value::from(orch.rules.merge.as_str()),
-                );
-                Self::save_toml_nested_value(
-                    "orchestrator",
-                    "rules",
-                    "review",
-                    toml_edit::Value::from(orch.rules.review.as_str()),
-                );
-                Self::save_toml_nested_value(
-                    "orchestrator",
-                    "rules",
-                    "custom",
-                    toml_edit::Value::from(orch.rules.custom.as_str()),
-                );
+                // Rewrite the whole `[orchestrator]` table in place. Previously this
+                // branch hand-wrote only `enabled`/`role`/`rules.*` and silently dropped
+                // every other field (notify/guardrails/auto_action_templates/
+                // pr_monitor_*) on save (#376). Serializing the full struct auto-covers
+                // future field additions and mirrors the per-project branch below.
+                let Some(path) = Self::config_path() else {
+                    tracing::debug!("No config path available, skipping save");
+                    return;
+                };
+                let content = std::fs::read_to_string(&path).unwrap_or_else(|e| {
+                    tracing::debug!(?path, %e, "Could not read config, starting fresh");
+                    String::new()
+                });
+                let mut doc = match content.parse::<toml_edit::DocumentMut>() {
+                    Ok(d) => d,
+                    Err(e) => {
+                        tracing::warn!(?path, %e, "Failed to parse config, starting fresh");
+                        toml_edit::DocumentMut::default()
+                    }
+                };
+
+                // Serialize via `{ orchestrator = <orch> }` wrapper so sub-tables are
+                // emitted as `[orchestrator.rules]`, `[orchestrator.notify]`, etc.
+                let mut wrapper = std::collections::BTreeMap::new();
+                wrapper.insert("orchestrator", orch);
+                let serialized = match toml::to_string(&wrapper) {
+                    Ok(s) => s,
+                    Err(e) => {
+                        tracing::warn!(%e, "Failed to serialize orchestrator settings; aborting save");
+                        return;
+                    }
+                };
+                let parsed = match serialized.parse::<toml_edit::DocumentMut>() {
+                    Ok(p) => p,
+                    Err(e) => {
+                        tracing::warn!(%e, "Failed to re-parse serialized orchestrator; aborting save");
+                        return;
+                    }
+                };
+                let Some(new_item) = parsed.as_table().get("orchestrator") else {
+                    tracing::warn!(
+                        "Serialized orchestrator missing [orchestrator] table; aborting save"
+                    );
+                    return;
+                };
+                // Assigning replaces the old `Item::Table` and all nested subtables,
+                // so stale entries from prior saves do not leak through.
+                doc["orchestrator"] = new_item.clone();
+
+                if let Some(parent) = path.parent() {
+                    if let Err(e) = std::fs::create_dir_all(parent) {
+                        tracing::warn!(?path, %e, "Failed to create config directory");
+                        return;
+                    }
+                }
+                if let Err(e) = std::fs::write(&path, doc.to_string()) {
+                    tracing::warn!(?path, %e, "Failed to write config file");
+                }
             }
             Some(proj_path) => {
                 // Save per-project orchestrator by rewriting [[project]] array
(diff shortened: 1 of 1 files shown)


trust-delta (commented):
LGTM — verified the fix and all critical checks pass.

## Verification summary

**Completeness** ✅ — `save_project_orchestrator_global_persists_all_fields` (settings.rs:2203) constructs `OrchestratorSettings` with non-default values across every nested group (`enabled`, `role`, `rules.*`, `notify.on_*`, `notify.buffer_*`, `notify.templates.*`, `guardrails.*`, `auto_action_templates.*`, `pr_monitor_*`) and asserts full equality after reload. Because `OrchestratorSettings` derives `PartialEq` (settings.rs:866) and any future field addition defaults into that comparison, the assertion auto-covers future regressions — matches the approach recommended in #376.

**No regression on per-project branch** ✅ — `save_project_orchestrator_per_project_still_works` (settings.rs:2285) exercises `Some(proj_path)`, confirms the target entry's override is written, and verifies a sibling `[[project]]` entry is preserved.

**Format preservation** ✅ — `save_project_orchestrator_global_preserves_unrelated_sections` (settings.rs:2255) seeds top-level scalar + `[ui]` and asserts both survive. The implementation only replaces `doc["orchestrator"]` on an existing `DocumentMut`, so other sections/comments are untouched. PR body correctly calls out that comments _inside_ the previous `[orchestrator]` block are lost on overwrite — matches per-project branch behaviour and is unavoidable.

**Idempotent / no stale-leak** ✅ — `save_project_orchestrator_global_overwrites_previous_save` (settings.rs:2221) saves rich values, then saves minimal values, and asserts that custom templates from the first save are gone. `doc["orchestrator"] = new_item.clone()` in `toml_edit` replaces the whole subtree, so orphaned subtables cannot linger.

**EventHandling as string** ✅ — `Serialize` impl at settings.rs:917-932 emits `"off" | "notify" | "auto_action"`. Round-trip tests confirm values survive.

**Serialization ordering** ✅ — verified `toml::to_string` reorders scalars before sub-tables automatically, so `pr_monitor_enabled` / `pr_monitor_interval_secs` (declared after the `auto_action_templates` sub-table) are correctly emitted at the top of `[orchestrator]` rather than leaking into `[orchestrator.auto_action_templates]`.

**Error handling** ✅ — every failure path (`config_path` missing, read fail, parse fail, serialize fail, re-parse fail, missing table, `create_dir_all`, `write`) uses `tracing::debug!` or `tracing::warn!` and returns. No panics.

**Scope** ✅ — only the `None` branch is rewritten. Struct, `save_toml_value`/`save_toml_nested_value`, and `src/web/api.rs` REST handler all untouched as required.

Ran `cargo test -p tmai-core --lib config::settings::` locally: 38 passed including the 4 new tests.

## Minor nit (non-blocking)

The new helpers (settings.rs:2182, 2229) use `let mut x = Default::default(); x.field = ...` which triggers `clippy::field_reassign_with_default`. The pattern is consistent with ~21 pre-existing instances in the crate, so this is not out of style — but if you run with `-D warnings` in the future, a struct-literal `OrchestratorSettings { enabled: true, role: ..., ..Default::default() }` would avoid the lint. Not a change request.

PR #397, opened 2026-04-12
fix: self-heal api.json when it disappears at runtime

trust-delta:
## Summary

- \`\$XDG_RUNTIME_DIR/tmai/api.json\` has been observed going missing multiple times while tmai's parent process is still alive. MCP clients then cannot reconnect (they read port+token from this file) even though the HTTP API keeps answering.
- Audit of every subcommand entrypoint (\`mcp\`, \`wrap\`, \`audit\`, \`codex-hook\`, \`init\`, \`uninit\`, \`tmux\`, \`webui\`) shows the **only** code path that deletes the file is the shutdown cleanup at \`src/main.rs:651\`, and that path was verified not to be reached during the most recent incident (current tmai process alive, single IPC-server-started line in the log, no shutdown messages).
- Root cause of the disappearance remains unidentified — no tmai.log trace, no concurrent process with delete access, no systemd tmpfiles cleanup, no external \`rm\` observed. Keeping the investigation open, but self-healing is safe and cheap.

## Fix

Add a tokio background task (\`mcp::client::spawn_api_info_watchdog\`) that polls once every 5s. If \`api.json\` is absent, it calls \`write_api_info\` with the same port+token the parent process is already using. Noop while the file is present.

Spawned from both \`run_webui_mode\` (\`src/main.rs:425\`) and \`run_tmux_mode\` (\`src/main.rs:212\`), right after the initial \`write_api_info\` call.

## Why not something stronger

Considered and rejected for now:
- **SIGHUP handler** — still requires user action, easy to forget.
- **IPC API for port/token** — solves the problem properly (MCP client would never need api.json at all) but is a larger architectural change; filed as a follow-up candidate.

## Test plan

- [x] \`cargo test -p tmai --lib mcp::client\` — 8 passed including new \`watchdog_rewrite_step_restores_missing_file\` (exercises the write step; the tokio interval itself is an integration concern not worth async test scaffolding).
- [x] \`cargo clippy --all-targets -- -D warnings\` — clean
- [x] \`cargo fmt --check\` — clean
- [ ] Manual: start tmai, delete api.json manually, confirm the file is recreated within ~5 seconds with the same port+token.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -209,6 +209,7 @@ async fn run_tmux_mode(settings: Settings, _cli: Config) -> Result<()> {
         if let Err(e) = tmai::mcp::client::write_api_info(settings.web.port, &token) {
             eprintln!("tmai: warning: failed to write API info: {e}");
         }
+        tmai::mcp::client::spawn_api_info_watchdog(settings.web.port, token.clone());
     }
 
     // Start orchestrator notifier if enabled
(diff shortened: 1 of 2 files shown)


trust-delta:
@coderabbitai review

PR #401, opened 2026-04-12
fix: stop useWorktrees from refreshing on every agents SSE tick

trust-delta:
## Summary

- Diagnosed today via Chrome DevTools MCP while chasing the \"input box unresponsive\" report: \`/api/worktrees\` was still being fetched dozens of times per second after #400 landed, because the remaining \`onAgents: scheduleRefresh\` path was triggering a refresh every time the backend emitted an \`agents\` SSE event.
- The backend emits \`agents\` on every Claude Code hook (PreToolUse, PostToolUse, UserPromptSubmit, Notification, Stop, …), so during active orchestrator use the 250ms debounce window never drained and the fetches piled up, swamping the input pipeline.
- Drop \`onAgents\` from useWorktrees entirely. Worktree list only changes via \`worktree_created\` / \`worktree_removed\`, which the backend already emits. The hook now refreshes on mount plus those two explicit events — no debounce needed.

## Related

- #400 — introduced the SSEProvider + initial 250ms debounce; this is the follow-up that makes the debounce unnecessary.
- Original symptom: \"input欄のレスポンスが悪い\" (2026-04-12 session)

## Test plan

- [x] \`npm run build\` + \`cargo build --bin tmai\` — binary rebuilt with new assets.
- [x] \`biome check\` / \`cargo fmt --check\` — clean.
- [ ] Manual: restart tmai, observe \`/api/worktrees\` fires only on mount + when worktree_created/removed SSE event lands.

## リリースノート

* **パフォーマンス改善**
  * ワークツリー機能の自動更新処理を最適化し、システムの応答性を向上させました。

* **UI改善**
  * マージ済みブランチの削除確認プロンプトのメッセージフォーマットを改善しました。


trust-delta:
Closing — misdiagnosed. The real report is typing-to-WebUI feedback lag during prompt composition (passthrough input), not /api/worktrees fetch pressure. Reopening the investigation on the passthrough + preview fetch path.

PR #416, opened 2026-04-12
feat: queue orchestrator notifications while operator is typing (#399)

### LINKED ISSUE #399: feat: queue orchestrator notifications while operator is typing
## Problem

While the operator is actively typing into the orchestrator agent's Claude Code input (but hasn't hit Enter yet), incoming \`OrchestratorNotifier\` prompts can be injected into the same input buffer via \`send_text\`. The half-typed user input gets clobbered or appended to the auto-injection, causing lost keystrokes, corrupted prompts, and in some cases the auto-prompt being submitted together with partial human input.

Observed in the 2026-04-12 session during manual orchestrator typing while PR Monitor events arrived.

## Current behaviour

The existing busy-buffer (PR #374) decides buffer-vs-deliver purely on agent \`phase\`:
- \`phase == Idle\` → deliver immediately
- \`phase != Idle\` → buffer, flush on transition to Idle

\"Typing but not yet submitted\" is indistinguishable from \"idle\" — \`phase\` stays \`Idle\` while the operator composes a prompt, so notifications fire straight into the buffer.

## Desired behaviour

While the operator has an in-flight composition (input buffer non-empty but no Enter yet), notifications should be queued and released only once the buffer goes empty OR the operator submits.

## Candidate signals for detecting \"typing\"

1. **Recent keystroke activity** — track the last time input was sent to the PTY (both human-originated and agent-originated). Treat any pane that received a keystroke in the last N seconds (default ~5s?) as \"typing\" → buffer.
2. **Input buffer contents from capture-pane** — the Claude Code prompt area sits on a known line; if the \`❯\` row has text after the cursor, there's pending composition. Requires parsing, fragile.
3. **Tmux activity timestamp** — \`display-message -p '#{pane_activity}'\` gives seconds-since-last-activity. Easy to read, but also fires on agent output, so needs pane-side origin discrimination.
4. **Hook-based UserPromptSubmit deadline** — the operator pressing Enter fires UserPromptSubmit. Until then, anything typed stays local. Need to distinguish human vs tmai-injected presses.

Option 1 feels cleanest: the CommandSender / send_text path already knows when a keystroke is pushed, so we can stamp a \"last human input\" timestamp per pane. The notifier consults this before deciding deliver-vs-buffer. The only subtlety is distinguishing human-initiated from tmai-initiated sends — tmai-initiated ones shouldn't mark the pane as \"operator is typing\".

## Scope

- \`crates/tmai-core/src/command_sender.rs\` — add an \`origin: ActionOrigin\` parameter (already exists) and only stamp the \"typing\" marker for \`ActionOrigin::Human\`.
- \`crates/tmai-core/src/state/store.rs\` — add \`last_human_input_at: Option<Instant>\` per-agent field.
- \`crates/tmai-core/src/orchestrator_notify/service.rs\` — extend the busy-or-idle check to also buffer if \`last_human_input_at\` is within the grace window (configurable, default 5s).
- \`crates/tmai-core/src/orchestrator_notify/buffer.rs\` — flush trigger already fires on \`AgentStatusChanged\`; we also need a \"grace window expired\" flush. Cheapest: piggyback on next \`AgentAppeared\` or a periodic tick.

## Tests

- Typing window active → notification buffered instead of delivered
- Typing window expires → buffered notification delivered on next tick or status change
- Agent-originated keystroke → does not stamp typing marker
- Immediate delivery when agent has never received human input (baseline case preserved)

## Priority

Medium — visible quality-of-life issue during active dogfooding but not a data loss / correctness problem. File it now so we can plan the work after the current PR queue drains.

## Related

- #372 / #374 — existing busy-buffer foundation (this builds on it)
- #384 — send_prompt Enter fix (orthogonal, but touches the same send-path)

trust-delta:
## Summary

- Track per-agent `last_human_input_at: Option<Instant>` on `MonitoredAgent`; stamp it only when `CommandSender` receives an `ActionOrigin::Human` keystroke, so WebUI passthrough marks the pane as "operator is composing" while agent/system sends stay silent.
- Extend the `OrchestratorNotifier` busy-check so an idle orchestrator is still treated as busy when `is_operator_typing(grace)` returns true — notifications queue into the existing `#372` busy-buffer instead of clobbering half-typed input.
- Add a 1-second periodic tick inside the notifier that flushes buffered entries once the typing grace expires (event-driven flush alone can't see "operator stopped typing without submitting"). Grace window is a configurable `typing_grace_secs` on `OrchestratorNotifySettings` (default 5s).

Resolves #399. Builds on #372/#374's busy-buffer foundation without redesigning it.

## Test plan

- [x] `cargo test -p tmai-core --lib command_sender` — Human origin stamps, Agent/System origins don't, tmux-target fallback works
- [x] `cargo test -p tmai-core --lib orchestrator_notify` — typing window buffers; idle-with-no-stamp still delivers immediately; periodic tick flushes once grace expires
- [x] `cargo fmt --all` + `cargo clippy` clean across the workspace
- [ ] Manual dogfood: type into orchestrator pane mid-composition while PR Monitor emits events; confirm the buffered prompt lands only after Enter or grace expiry

## リリースノート

* **新機能**
  * ユーザー入力の検出機能を追加しました。
  * 入力検出の猶予時間を設定できるようになりました。デフォルトは5秒です。
  * ユーザー入力中は、システム通知の送信を遅延させるようになりました。

### DIFF
--- a/crates/tmai-core/src/agents/types.rs
+++ b/crates/tmai-core/src/agents/types.rs
@@ -810,6 +810,11 @@ pub struct MonitoredAgent {
     pub issue_number: Option<u64>,
     /// Associated PR number (auto-detected from branch or set manually)
     pub pr_number: Option<u64>,
+    /// Timestamp of the most recent human-originated keystroke into this
+    /// agent's pane. Used by `OrchestratorNotifier` to defer notifications
+    /// while the operator is composing input but has not submitted yet (#399).
+    /// Not serialized — purely an in-memory signal.
+    pub last_human_input_at: Option<std::time::Instant>,
 }
 
 impl MonitoredAgent {
@@ -880,7 +885,23 @@ impl MonitoredAgent {
             is_orchestrator: false,
             issue_number: None,
             pr_number: None,
+            last_human_input_at: None,
         }
+    }
+
+    /// Stamp `last_human_input_at` with the current time. Called from
+    /// `CommandSender` when a keystroke originates from `ActionOrigin::Human`.
+    pub fn note_human_input(&mut self) {
+        self.last_human_input_at = Some(std::time::Instant::now());
+    }
+
+    /// Returns true when the operator is actively composing input into this
+    /// pane (i.e. a Human keystroke landed within the last `grace` window).
+    /// Used by the orchestrator notifier to avoid clobbering in-flight input.
+    pub fn is_operator_typing(&self, grace: std::time::Duration) -> bool {
+        self.last_human_input_at
+            .map(|t| t.elapsed() < grace)
+            .unwrap_or(false)
     }
 
     /// Set the detection source
--- a/crates/tmai-core/src/command_sender.rs
+++ b/crates/tmai-core/src/command_sender.rs
@@ -2,6 +2,7 @@ use anyhow::Result;
 use std::collections::HashMap;
 use std::sync::Arc;
 
+use crate::api::ActionOrigin;
 use crate::codex_ws::CodexWsSender;
 use crate::hooks::registry::HookRegistry;
 use crate::ipc::server::IpcServer;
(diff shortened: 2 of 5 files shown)


trust-delta (commented):
LGTM overall. Core design is sound: stamping is gated correctly on `ActionOrigin::Human`, the two flush paths (status-change and periodic tick) dedupe atomically via `buffer::take_for_flush`'s `map.remove`, and the busy-buffer path from #372/#374 is preserved (the new `!is_operator_typing(grace)` condition only narrows idle-eligibility, pushing deferred notifications into the existing busy branch rather than forking control flow).

A few observations, none blocking:

**1. Stamping coverage is WebUI-only** — `src/ui/app.rs` has ~12 call sites (e.g., lines 883, 892, 908, 986, 991, 1926, 1950) that send human keystrokes via the non-origin `send_keys` / `send_keys_literal` methods. `ActionOrigin::tui()` already exists but isn't used. The PR description explicitly scopes to WebUI passthrough, so this is a natural follow-up rather than a gap in this PR — but worth tracking: a TUI operator composing in the input-mode popup won't get typing-grace protection today.

**2. Both flush paths are needed and coexist cleanly** — `flush_expired_typing_graces` (tick) handles the "operator stops typing without submitting" case that event-driven flush can't see, while `maybe_flush_for_event` handles the common status→idle transition. Since both run in the same tokio task under a biased `select!`, they can't execute concurrently; `take_for_flush`'s atomic `map.remove` means no double-delivery even if both paths target the same entry across ticks. 👍

**3. `typing_grace_secs = 5` default** — reasonable for the common WebUI streaming case where each char resets the window. One subtle implication: `Enter` is itself a Human keystroke, so submitting also stamps. That means buffered notifications to an orchestrator that's now processing the submitted prompt wait for (a) Busy→Idle AND (b) ≥5s since Enter. Correct — the operator likely wants to read the response first — but worth a doc-comment note that post-submit delivery carries an implicit grace delay.

**4. `test_typing_grace_expiry_flushes_buffered_entry` decomposes rather than integrates** (orchestrator_notify/service.rs:697-767) — it pre-seeds the buffer with `buf.write().entry(...)` instead of driving a buffered entry through the `handle_event` → "typing→buffered" path. The three tests together cover each branch, but a single end-to-end test (stamp typing → fire event → assert buffered → advance time / clear stamp → assert tick flush) would catch wiring bugs in how the two paths compose. Not a blocker.

**5. `maybe_stamp_human_input` fallback loop** (command_sender.rs:112-117) — takes a write lock and O(n) scans `state.agents` when the map-key lookup misses. Benign with parking_lot and typical agent counts, but if it becomes hot under keystroke streaming, consider caching a tmux-target→map-key reverse index on `AppState`.

**6. Doc-comment accuracy** — the "Not serialized" note on `last_human_input_at` (agents/types.rs:811-814) is vacuously true: `MonitoredAgent` doesn't derive `Serialize`/`Deserialize`. Fine as-is; just flagging that the field's `Instant` type would break a future `Serialize` derive without a `#[serde(skip)]` marker.

PR #418, opened 2026-04-12
bug: parallel dispatch_issue misattributes issue_number metadata across agents (#414)

### LINKED ISSUE #414: bug: parallel dispatch_issue misattributes issue_number metadata across agents
## Observed

When two `dispatch_issue` calls run concurrently (via MCP), agent → issue_number metadata ends up crossed.

### Repro (2026-04-13 orchestrator session)

Two near-simultaneous MCP calls from the same orchestrator:
\`\`\`
dispatch_issue(issue_number=399) → { session_id: \"session-1:2.2\" }
dispatch_issue(issue_number=413) → { session_id: \"session-1:2.2\" }   # ← identical
\`\`\`

Actual placement after spawn settled:
| pane | branch / cwd                                      | issue_number in metadata |
|------|---------------------------------------------------|--------------------------|
| 2.2  | 413-perf-preview-panel-re-renders-...             | (none)                   |
| 2.3  | 399-feat-queue-orchestrator-notifications-...     | **413** (wrong)          |

Expected: pane 2.2 → 413, pane 2.3 → 399.

Orchestrator pane itself also shows `pr_number: 412` (the PR that was just merged in this session) despite being the main-repo orchestrator, not a worktree agent — possibly a related pr-attribution path.

## Likely root cause

`dispatch_issue_handler` (`src/web/api.rs:3374-3393`) flow:
1. Call `spawn_in_tmux`, receive `session_id`
2. Either mutate existing agent in state, or insert into `pending_agent_metadata` keyed by `session_id`
3. Poller later detects the tmux pane, applies pending metadata (`crates/tmai-core/src/state/store.rs:874`)

Two failure modes possible (not yet confirmed which dominates):

**(A) tmux spawn returns the same placeholder session_id for concurrent calls.** The MCP response for both dispatches showing `session-1:2.2` is direct evidence. #337 (PR #345 \"serialize tmux spawns to prevent duplicate session_id\") serializes PTY spawns, and its regression test (`concurrent_pty_spawns_return_unique_session_ids`, `src/web/api.rs:5129`) covers `spawn_in_pty` only. The tmux path in `spawn_in_tmux` may not share the same serialization, leaving a window where two concurrent calls compute the same target pane before tmux actually allocates them.

**(B) pending_agent_metadata is keyed by the wrong session_id.** If call #1 writes `meta[\"session-1:2.2\"] = {issue: 399}` and call #2 overwrites with `meta[\"session-1:2.2\"] = {issue: 413}`, then when the poller detects the **399 agent** on pane `session-1:2.3` it looks up `meta[\"session-1:2.3\"]` → empty. The **413 agent** lands on `session-1:2.2` and correctly picks up `issue=413`. Observation contradicts this (399 agent got 413, 413 agent got none), so mode A (wrong session_id returned from spawn) is more likely — the returned session_id was stale/guessed, actual pane allocation went the other way, and the metadata key does not match the real landing pane.

## Reproduction plan

1. Open a test that runs two `dispatch_issue_handler` invocations concurrently against `spawn_in_tmux` (mock or real)
2. Assert returned session_ids are distinct
3. Assert after poller tick, each agent's `issue_number` matches its worktree branch

The existing PTY-focused `concurrent_pty_spawns_return_unique_session_ids` test is a template; we need a `concurrent_tmux_spawns_return_unique_session_ids` equivalent plus a higher-level dispatch_issue concurrency test that verifies metadata attribution.

## Fix direction

- Apply the same serialization #337 used for PTY to the tmux spawn path (if not already), so concurrent spawns cannot observe the same session_id.
- Alternative / complementary: key `pending_agent_metadata` on something more stable than the spawn-returned session_id (e.g., worktree path or branch name), which the poller already knows from the agent's `cwd`/`git_branch`.

## Priority

Medium-high — silently misroutes orchestrator actions (auto_cleanup, PR-based lookups, future semantic addressing) in any multi-agent workflow, and is easy to trigger whenever an orchestrator fans out.

## Related

- #331 — original semantic addressing work (closed; this is a regression in a narrower case)
- #337 / PR #345 — duplicate session_id fix for PTY path

trust-delta:
## Summary
- Root cause: `pending_agent_metadata` was keyed by the spawn-returned `session_id`. When two `dispatch_issue`/`dispatch_review` calls ran concurrently and the spawn layer ever reported a duplicate or stale pane target, both writes collided at the same key — so the pane that actually ran a given worktree could receive wrong or empty `issue_number`/`pr_number`.
- Fix: rekey `pending_agent_metadata` by worktree path (`agent.cwd`), which is unique per dispatch and stable across pane renumbering. Poller applies metadata by matching `agent.cwd` on first detection. Obsolete pane-renumber migration for this map is dropped.
- Added regression test `test_pending_metadata_keyed_by_worktree_path_attributes_correctly` in `state::store` reproducing the repro scenario: two pending entries, pane landings crossed relative to dispatch order — both agents get correct attribution after `update_agents`.

## Test plan
- [x] Failing test reproduces the bug on main (verified before fix)
- [x] Test passes after the fix
- [x] `cargo test -p tmai-core --lib` (996 passed; 1 pre-existing unrelated failure in `detectors::claude_code`)
- [x] `cargo test -p tmai --lib` (141 passed)
- [x] `cargo fmt --all`, `cargo clippy` (no new warnings)

Fixes #414.

### DIFF
--- a/crates/tmai-core/src/state/store.rs
+++ b/crates/tmai-core/src/state/store.rs
@@ -539,7 +539,10 @@ pub struct AppState {
     pub pending_orchestrator_ids: HashSet<String>,
 
     /// Pending metadata for agents not yet detected by the poller.
-    /// Key: session_id from spawn response. Applied and removed on first agent detection.
+    /// Key: worktree path (agent `cwd`). Keyed by a stable path rather than
+    /// the spawn-returned `session_id` so that concurrent dispatches cannot
+    /// cross-wire attribution if the spawn layer ever reports a duplicate or
+    /// stale target (#414). Applied and removed on first agent detection.
     /// Used by dispatch_issue (issue_number) and dispatch_review (pr_number).
     pub pending_agent_metadata: HashMap<String, PendingAgentMetadata>,
 
@@ -767,10 +770,8 @@ impl AppState {
                     if self.pending_orchestrator_ids.remove(old_target) {
                         self.pending_orchestrator_ids.insert(new_target.clone());
                     }
-                    // Migrate pending_agent_metadata if keyed by old target
-                    if let Some(meta) = self.pending_agent_metadata.remove(old_target) {
-                        self.pending_agent_metadata.insert(new_target.clone(), meta);
-                    }
+                    // pending_agent_metadata is keyed by worktree path
+                    // (stable across pane renumbering, #414) — no migration.
                     // Update target_to_pane_id mapping
                     if let Some(pane_id) = self.target_to_pane_id.remove(old_target) {
                         self.target_to_pane_id.insert(new_target.clone(), pane_id);
@@ -870,8 +871,10 @@ impl AppState {
                 if self.pending_orchestrator_ids.remove(&id) {
                     agent.is_orchestrator = true;
                 }
-                // Apply pending metadata (issue_number, pr_number, base_branch)
-                if let Some(meta) = self.pending_agent_metadata.remove(&id) {
+                // Apply pending metadata (issue_number, pr_number, base_branch).
+                // Keyed by worktree path (agent.cwd) — stable across pane
+                // renumbering and robust to duplicate spawn session_ids (#414).
+                if let Some(meta) = self.pending_agent_metadata.remove(&agent.cwd) {
                     if meta.issue_number.is_some() {
                         agent.issue_number = meta.issue_number;
                     }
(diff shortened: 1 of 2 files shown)


trust-delta (commented):
## Review summary

The fix is well-scoped and addresses the core race correctly: keying `pending_agent_metadata` by the stable worktree path instead of the spawn-returned `session_id` eliminates cross-wire attribution when the tmux spawn layer reports a duplicate/stale target. All three sites move in lockstep (writer in `spawn_worktree`, writer in `perform_dispatch_review`, reader in `AppState::update_agents`), and the corresponding pane-renumber migration in `update_agents` is correctly dropped since the new key is invariant under pane renumbering. LGTM to merge.

A few concerns worth tracking — none are merge blockers:

### 1. Duplicate-`session_id` fast path is still keyed by session_id
`src/web/api.rs:3211` (dispatch_review) and `src/web/api.rs:3386` (spawn_worktree) still take the "agent already in state" branch via `s.agents.get_mut(&resp.session_id)`. If the spawn layer returns a stale/duplicate `session_id` that collides with a *different* already-detected agent, metadata is written onto the wrong agent record directly — the new pending-path routing does not cover this. In the #414 scenario the pending path is what fires (freshly spawned tmux pane not yet in state), so practical exposure is low, but the two code paths now disagree about which key is trustworthy. Consider either (a) also resolving the fast path by `agent.cwd == worktree_path` or (b) a brief comment acknowledging the residual trust in `session_id` for the already-detected case.

### 2. Regression test validates the new scheme but does not reproduce the original failure
`test_pending_metadata_keyed_by_worktree_path_attributes_correctly` (`crates/tmai-core/src/state/store.rs:2284`) inserts directly into `pending_agent_metadata` with worktree-path keys and checks that `update_agents` applies them by `agent.cwd`. It would fail under the old code (which looked up by `&id`/pane target), so it guards the new implementation — but it does not model the *actual* bug, which was two concurrent writers colliding on the same `session_id` key and overwriting each other. A test that, under the old keying scheme, inserts two entries with the same key and demonstrates attribution loss would be a more faithful repro. Additionally, there is no HTTP-layer concurrent-spawn test analogous to `concurrent_pty_spawns_return_unique_session_ids` at `src/web/api.rs:5135` — a parallel pair of `spawn_worktree`/dispatch_issue calls driven through the router would tighten coverage end-to-end.

### 3. Reliance on byte-exact `agent.cwd` match
Poller populates `agent.cwd` from tmux's `#{pane_current_path}` (`crates/tmai-core/src/tmux/client.rs:87`), which resolves via `/proc/<pid>/cwd` and is canonical. The writer uses `wt_result.path` from `git worktree add` output. If those ever diverge in trailing-slash handling or symlink resolution (e.g. `/home/user` vs a canonical path on macOS where `/tmp` → `/private/tmp`), the lookup silently misses and metadata is dropped with no observable error. Not a problem on today's Linux-first target, but worth a canonicalization step on both sides, or at minimum a `tracing::debug!` on the miss branch so future drift is detectable.

### 4. Backwards compat
All three existing writers (`spawn_worktree` covering `dispatch_issue`, `perform_dispatch_review`, the removed pane-renumber migration) are consistent with the new key. No orphaned writers remain.

PR #419, opened 2026-04-12
perf: preview panel re-renders entire scrollback on every tick (input lag in long sessions) (#413)

### LINKED ISSUE #413: perf: preview panel re-renders entire scrollback on every tick (input lag in long sessions)
## Problem

In long-running Claude Code sessions, typing into the preview panel becomes laggy because every preview poll tick re-parses and re-writes the entire Hybrid Scrollback content (hundreds of KB up to several MB) through AnsiUp → DOMPurify → innerHTML.

Evidence from `crates/tmai-app/web/src/components/agent/PreviewPanel.tsx`:
- Inline comment: "Preview responses can be hundreds of KB to several MB (Hybrid Scrollback), and feeding an identical string through setContent → AnsiUp → DOMPurify → innerHTML on every 100ms active-input tick was the main cause of input lag."
- Another comment: "rewriting the entire innerHTML (1.4MB in an active session) unmounts the hidden <input>..."

Byte-equal dedupe (`lastContentRef` / `lastHtmlRef`) mitigates idle ticks, but as soon as the agent emits any output the full blob is re-rendered even though >99% of it (the scrollback history) is immutable.

## Hypothesis

The `content` payload returned by `GET /api/preview` mixes two logically different regions:
1. **Immutable scrollback history** — past JSONL-rendered conversation, never changes once written
2. **Live capture-pane tail** — the bottom few hundred lines that actually update each tick

Rendering them as one string forces unnecessary work. A `TranscriptView` component already renders JSONL transcript records separately — so the history layer inside `content` is effectively duplicated work in that case.

## Proposed direction

**Option A (cleanest): drop scrollback history from `content`.** Let `TranscriptView` own the history layer (it already does), and have the backend return only the live capture-pane in `content`. The live portion is small (single-screen order), so re-render cost drops by 1-2 orders of magnitude.

**Option B (fallback): frontend diff.** Detect prefix-match between new and old `content`; when the new payload extends the old, only append the new suffix to the DOM instead of rewriting innerHTML. Preserves backend contract.

**Option C: split payload.** Backend returns `{ history, live }`; frontend renders history once (or when it grows), live on every tick.

Option A depends on whether TranscriptView coverage is complete vs. the scrollback (agents without hook connection have no transcript). Option B is backend-neutral but fragile around ANSI boundaries. Option C is the most explicit contract.

## Notes

- Only visible in long sessions — short sessions don't accumulate enough scrollback to notice
- Especially noticeable when composing IME input; even though `fetchPreview` skips DOM updates during `composing`, post-keystroke two-stage fetch (50ms + 200ms) still rewrites the whole thing
- User reports the symptom disappears when a fresh Claude Code session is started, consistent with accumulated-history cost
- Contributing factor alongside Claude Code's own conversation-length rendering cost, but addressable on tmai side

## Acceptance

- Re-render cost during active typing scales with live-capture size, not total scrollback size
- Verified on a session with >1MB of scrollback history: keystroke → visible update latency unchanged as history grows
- No regression in cursor overlay accuracy or text selection

trust-delta:
## Summary
- Backend `/api/preview` now returns `live_start_line`, so the client can slice the tmux capture-pane output at the `history_size` boundary instead of re-parsing the full blob on every poll tick.
- `PreviewPanel` renders scrollback history (memoized on its string value) into its own `<div>` and the visible region into a second `<div>`. AnsiUp + DOMPurify + innerHTML only run on history when scrollback actually grows; live (~one screenful) keeps updating every tick. Cursor marker stays in live.
- Falls back to treating the whole content as live when the runtime can't report pane dimensions (standalone runtime, last_content cache), so the legacy `web/` UI is unaffected.

Closes #413.

## Why option C and not A
TranscriptView is only populated when the hook connection is present, so dropping scrollback from `content` (Option A) would leave agents without hooks blind. Option C — backend splits `{content, live_start_line}` — works uniformly.

## Test plan
- [ ] `pnpm test` — 8 new cases in `preview-content.test.ts` cover splitting at line boundaries, edge cases (empty content, over-large `liveStartLine`), and verify that live-only render is ≥10× faster than full-content render on a ~1.2 MB history fixture.
- [ ] `pnpm run build` — regenerates `src/web/assets/`.
- [ ] `pnpm run lint` — biome clean.
- [ ] `cargo build` + `cargo test --lib -p tmai-core tmux` — tmux client tests pass.
- [ ] Manual: open a long Claude Code session (>1 MB scrollback) and type in the preview panel; keystroke→echo latency should be independent of scrollback size. Verify cursor overlay still lands on the live region and text selection inside history still works.


trust-delta (commented):
## Summary

Option C (`{content, live_start_line}`) is a clean choice — history/live boundary is computed once at the source of truth (tmux `history_size`) and the client just splits on the Nth newline. The design correctly keeps the cursor on the live slice and falls back to whole-content-as-live when pane dims aren't available. LGTM, with a few small notes below.

### 1. `live_start_line` correctness

Consistent: backend uses `min(history_size, lines)` and the JS helper finds the Nth `\n` (history = `[0, pos)`, live = `[pos+1, ...)`). Typical invariant `lines == history_size + pane_height` holds, so the split lands exactly on the first visible line.

Edge cases I walked through:
- **Empty history / `history_size=0`** → `live_start=0`, `splitPreviewContent` short-circuits → whole content is live. ✓
- **Capture smaller than expected** (`lines < history_size`): backend clamps with `.min(lines)`. In the extreme where `live_start == lines` with a trailing `\n`, `splitPreviewContent` locates the Nth newline at the end, yielding `live = ""`. Not a crash — history renders, cursor overlay is hidden — but live is visually empty. Worth a one-line guard if you want to be defensive: if `pos === content.length - 1`, fall back to whole-content-as-live. Practically unreachable unless `pane_height == 0`.
- **Compact/scrollback reset**: `history` string changes → memo recomputes → HTML re-written. ✓
- **Race between `get_pane_view_info` and `capture_pane_full`**: if new lines appear between the two calls, the stale (smaller) `history_size` puts a few scroll-off-bound lines into `live`. Harmless — they just re-render once. ✓

### 2. Standalone runtime fallback

`RuntimeAdapter::get_pane_view_info` defaults to `Ok(None)` → `pre_view = None` → `live_start_line = 0` → JS treats the whole payload as live. `history` stays `""`, `historyHtml` short-circuits to `""` and the `lastHistoryHtmlRef === ""` check blocks an empty `innerHTML` write. No blank render, no duplicate render of the blob. ✓

Minor note: `get_pane_view_info` is now called unconditionally (old code skipped `get_cursor_position` when `show_cursor=false`). One extra cheap `display-message` tmux call per tick in the no-cursor path — worth it for the split, but not free. Non-blocking.

### 3. Cursor marker placement

Cursor `y` coming back from `/api/preview` is absolute (`history_size + v.cursor_y`, clamped to `lines-1`). `liveCursor` subtracts `liveStartLine`, so the coordinates fed to the marker-injection loop in `liveHtml` are relative to the live slice — matching what `base.split("\n")` sees. `liveRef`'s marker `offsetTop` is resolved against the `.ansi-preview` positioned ancestor, which wraps both `historyRef` and `liveRef`, so the absolute pixel offset naturally includes the history block's height. ✓

If `cursor_y < liveStartLine` (race or clamp under-shoot), `relY < 0` → cursor hidden. Correct conservative behavior.

### 4. Memoization keys on the history `<div>`

`useMemo([ansi, history, cols])` uses `Object.is` — strings compare by value. A "false-equal" would require two byte-identical history strings, in which case AnsiUp output is also identical, so there's no hazard. The `setHistory(prev => prev === split.history ? prev : split.history)` guard bails setState when the value is unchanged, which cascades to the effect skipping the `innerHTML` write. Looks correct.

### 5. Perf claim

The `fullElapsed > liveElapsed * 10` assertion is methodologically fine as a floor — the ratio in practice is much higher, so CI jitter shouldn't flip it. Two small concerns though:
- Timing-based tests can still flake under load. A call-count test (spy on `ansi_to_html` and assert it isn't called when `history` is unchanged across ticks) would verify the actual invariant the fix is claiming, independent of wall-clock.
- `trimPreviewContent` is applied to the full content in the "full render" arm but only `shrinkContentToWidth` in the real history path — the benchmark comparison is apples-to-apples for AnsiUp cost but very slightly overstates the full-render cost by including the trailing-blank-line regex. Minor.

Not blocking — the perf win is real and the test surfaces an obvious regression if someone breaks the split.

### 6. Legacy `src/web/` UI

`src/web/assets/` is the built bundle of `crates/tmai-app/web/`; there is no separate legacy UI source. The rename `index-B0y4ly_U.js` → `index-Bgw4jfzp.js` in the diff is just the new bundle hash. Claim holds. ✓

### Other notes

- No Rust unit test covers the new 4-part `display-message` parse in `TmuxClient::get_pane_view_info`. The parsing is straightforward (and identical in shape to the old 3-part path), but a quick `test_parse_pane_view_format` would be cheap insurance against future format drift.
- The `// Clamp cursor_y to captured content bounds` comment was lost in the reshuffle in `src/web/api.rs`; the logic is still there (`absolute_y.min((lines - 1) as u32)`) but no comment. Minor.

Overall: nice fix, clear test coverage for the regression, safe degradation on the non-tmux path.

PR #420, opened 2026-04-12
feat: render mermaid diagrams in MarkdownPanel (#417)

### LINKED ISSUE #417: feat: render mermaid diagrams in MarkdownPanel
## Problem

The WebUI MarkdownPanel renders Markdown via `react-markdown` + `remark-gfm` only (`crates/tmai-app/web/src/components/markdown/MarkdownPanel.tsx:231`). ` ```mermaid ` fenced blocks pass through as plain code blocks, so architecture docs, decision records, and research notes that embed Mermaid are unreadable in-app.

Noticed while reviewing `.claude/research/2026-04-13_architecture-overview.md`, which contains 3 Mermaid diagrams.

## Proposed

Add client-side Mermaid rendering inside the `components` override of `react-markdown`:

- Install `mermaid` (npm)
- In `MarkdownPanel.tsx`, add a `components.code` override:
  - Detect `className === \"language-mermaid\"`
  - Render into a `<div>` ref and call `mermaid.render(id, source)` inside `useEffect`
  - Dark-theme config to match the existing `prose-invert` styling (`theme: 'dark'` or custom CSS vars)
  - Fallback to the original code block if render throws (so malformed diagrams are still readable)

### Implementation notes

- `mermaid` is ~500KB min+gzip. Dynamic import (`const mermaid = await import('mermaid')`) so it only loads on pages with a Mermaid block — the MarkdownPanel is not always mounted.
- Each diagram needs a stable unique id; use the source hash or a counter per panel mount.
- After dependency install, remember the build step: `cd crates/tmai-app/web && npm run build` to refresh `src/web/assets/`.

## Out of scope

- No server-side rendering to SVG (keeps the backend free of headless browser deps)
- No editable/interactive Mermaid — read-only render is enough

## Test plan

- Architecture doc renders all 3 diagrams without layout break
- Invalid Mermaid block falls back to code-block display
- No Mermaid in file → no extra bundle loaded (verify dynamic import fires only when needed)

## Priority

Medium — blocks reading architecture/design documents in-app. Low risk, bounded scope.

trust-delta:
Closes #417.

## Summary
- Dynamically load `mermaid` from a `MermaidBlock` component (only fetched when a rendered doc actually contains a ` ```mermaid ` fence)
- Wire `MermaidBlock` into `MarkdownPanel.tsx` via a react-markdown `pre` component override; non-mermaid code blocks are untouched
- Fall back to the original plain code block if `mermaid.render` throws, so a single malformed diagram cannot break the whole panel
- Use `theme: 'dark'` + `securityLevel: 'strict'` to match `prose-invert` styling and keep mermaid's DOMPurify sanitization

## Test plan
- [x] `pnpm run test` (96 passed, incl. new `extractCodeSource` helper tests)
- [x] `pnpm run lint`
- [x] `pnpm run build` — confirmed `mermaid.core` and every diagram-type (`flowDiagram`, `sequenceDiagram`, `c4Diagram`, …) land in separate lazy chunks; `src/web/assets/` regenerated
- [ ] Manual: open `.claude/research/2026-04-13_architecture-overview.md` in the WebUI and verify all 3 diagrams render, and a malformed ` ```mermaid ` block falls back to a code view (not exercised in this sandbox — backend-served file tree not available)

## Summary by CodeRabbit

# リリースノート

## 新機能
* **Mermaidダイアグラムサポート**: マークダウン内で Mermaid 図表記法を使用できるようになりました。コードブロックで `language-mermaid` を指定することで、フローチャート、ER図、状態図など複数のダイアグラムタイプをレンダリング可能です。

* **マークダウンパネル強化**: マークダウンレンダラーが Mermaid フェンスドコードブロックをインターセプトし、専用の `MermaidBlock` コンポーネント経由で動的にレンダリングするようになりました。


trust-delta (commented):
Nice, tight implementation — dynamic import is correctly scoped to `MermaidBlock` (only pulled when a ` ```mermaid ` fence is actually rendered), `securityLevel: 'strict'` is preserved with no `htmlLabels` override, and the per-block error boundary in `MermaidBlock` keeps a malformed diagram from nuking the rest of the panel. A few observations, mostly non-blocking:

### Potentially worth addressing

**`mermaidPromise` caches rejection forever** — `MermaidBlock.tsx:9,13-27`
If the initial `import("mermaid")` fails (chunk 404, offline blip, CSP rejection), `mermaidPromise` is left pointing at the rejected promise. Every subsequent `MermaidBlock` mount on that page session will hit `.then(...)`'s rejection path and fall back to a plain code block — no retry until full page reload. Consider nulling the cache on rejection so a later mount can retry:

```ts
mermaidPromise = import("mermaid").then(...).catch((e) => {
  mermaidPromise = null;
  throw e;
});
```

**Test mirrors the helper instead of importing it** — `__tests__/mermaid-source.test.ts:6-10`
`extractCodeSource` is re-declared in the test file with a comment saying this matches "the existing component helper tests in this repo". This means a future refactor of the real `extractCodeSource` in `MarkdownPanel.tsx:18-22` won't be caught — the test verifies a copy. If the repo convention really is to duplicate, fine; otherwise it's worth exporting the helper and importing it in the test.

**No smoke test for the actual component mount** — given the PR description flags "not exercised in sandbox" as an unchecked manual item, a vitest + jsdom test that renders `<MermaidBlock source="flowchart TD\n  A --> B" />` with a mocked `import("mermaid")` would catch the mount path in CI. The current build step catches bundle regressions but not mount failures.

### Minor / stylistic

- **`instanceCounter` double-increment** — `MermaidBlock.tsx:33,39`: initial `useRef` pre-increments (`mermaid-1`), then useEffect immediately overwrites with `mermaid-2`. Works, but the initial `useRef` value is dead. `useRef<string>("")` would read cleaner.
- **Silent parse-error fallback** — `MermaidBlock.tsx:47-51`: on malformed diagrams the user just sees a code block with no clue why. A `console.warn("[MermaidBlock]", e)` before `setError` would help developers debug without affecting UX.
- **Dark theme / `prose-invert`** — `theme: 'dark'` is the right call. Visual alignment with the exact `prose-invert` palette (zinc-100/300 text, white/5 borders) is a QA pass you can only do in the running WebUI; worth noting in the follow-up manual test when the backend is available.

### Good stuff
- Dynamic import is truly lazy — `loadMermaid()` is only invoked from `MermaidBlock`'s useEffect, which is only rendered through the `language-mermaid` branch in `MarkdownPanel.tsx:29-32`. No eager import into the entry bundle.
- `securityLevel: 'strict'` + mermaid's internal DOMPurify is the right stance, and the `biome-ignore` comment on `dangerouslySetInnerHTML` is correctly justified.
- Per-block catch boundary is tight and only swallows render errors from the current block's mermaid call.

Not blocking merge — the rejection-cache issue is the only real bug and is low-frequency in practice.

PR #429, opened 2026-04-13
bug: Rule Preset 'tmai MCP tools' toggle does not persist (allow_tmai_mcp dropped on PUT) (#428)

### LINKED ISSUE #428: bug: Rule Preset 'tmai MCP tools' toggle does not persist (allow_tmai_mcp dropped on PUT)
## Symptom

Toggling **Settings → Rule Presets → tmai MCP tools** has no effect — the switch always reverts to its previous state on next render. Other Rule Preset toggles work normally.

## Root cause

Two backend struct definitions in \`src/web/api.rs\` are missing the \`allow_tmai_mcp\` field:

1. **PUT request body** — \`UpdateRuleSettingsRequest\` (\`src/web/api.rs:1897-1904\`) only declares \`allow_read\`, \`allow_tests\`, \`allow_fetch\`, \`allow_git_readonly\`, \`allow_format_lint\`, \`allow_patterns\`. The frontend sends \`{ rules: { allow_tmai_mcp: true } }\` (\`crates/tmai-app/web/src/lib/api-http.ts:1014, 570\`), serde silently drops the unknown field, and the persistence branch never fires.

2. **GET response body** — \`RuleSettingsResponse\` returned by \`get_auto_approve_settings\` (\`src/web/api.rs:1917-1924\`) also omits the field, so the UI's hydrated state never reflects what's in \`config.toml\` for \`allow_tmai_mcp\`.

Underlying setting (\`crates/tmai-core/src/config/settings.rs:621-623\`) and rules engine (\`crates/tmai-core/src/auto_approve/rules.rs:315-316, 444-445\`) handle the field correctly — only the WebUI bridge is broken.

## Fix

- Add \`pub allow_tmai_mcp: Option<bool>\` to \`UpdateRuleSettingsRequest\` and the matching persistence branch using \`save_toml_nested_value(\"auto_approve\", \"rules\", \"allow_tmai_mcp\", ...)\`
- Add \`pub allow_tmai_mcp: bool\` to \`RuleSettingsResponse\` and populate from \`aa.rules.allow_tmai_mcp\`

Both edits are 1-2 lines each. After the fix, run \`cd crates/tmai-app/web && pnpm run build\` to refresh \`src/web/assets/\` (no frontend code change needed — schema already includes the field).

## Test plan

- New unit test on \`update_auto_approve_settings\` that PUTs \`{ rules: { allow_tmai_mcp: false } }\` and asserts the saved toml contains \`allow_tmai_mcp = false\`
- GET test that asserts \`allow_tmai_mcp\` round-trips

## Priority

P1 — silent setting failure breaks user trust in the panel; default is \`true\` so users explicitly turning it OFF for safety see no effect.

## Related

- This fits the SoT-unification theme but is narrower: it's a missing field, not a duplication. File now, fix immediately.

trust-delta:
## Summary

- Add `allow_tmai_mcp` to `UpdateRuleSettingsRequest` + persistence branch in `update_auto_approve_settings` so PUT `/api/settings/auto-approve` no longer silently drops the field.
- Add `allow_tmai_mcp` to `RuleSettingsResponse` and populate it from `aa.rules.allow_tmai_mcp` in `get_auto_approve_settings` so GET round-trips the persisted value.
- Regression unit tests: PUT body deserialization captures `allow_tmai_mcp`, GET response reflects the settings value.

Fixes #428.

## Test plan

- [x] `cargo test --package tmai --lib -- web::api::tests::test_update_rule_settings_accepts_allow_tmai_mcp web::api::tests::test_get_auto_approve_returns_allow_tmai_mcp`
- [x] `cargo fmt --all` / `cargo clippy --all-targets -- -D warnings`
- [x] `pnpm run build` in `crates/tmai-app/web` (no frontend diff — schema already had the field)
- [ ] Manual: toggle **Settings → Rule Presets → tmai MCP tools** off, reload — the switch stays off and `~/.config/tmai/config.toml` shows `allow_tmai_mcp = false`.

### DIFF
--- a/src/web/api.rs
+++ b/src/web/api.rs
@@ -1854,6 +1854,7 @@ pub struct RuleSettingsResponse {
     pub allow_fetch: bool,
     pub allow_git_readonly: bool,
     pub allow_format_lint: bool,
+    pub allow_tmai_mcp: bool,
     pub allow_patterns: Vec<String>,
 }
 
@@ -1900,6 +1901,7 @@ pub struct UpdateRuleSettingsRequest {
     pub allow_fetch: Option<bool>,
     pub allow_git_readonly: Option<bool>,
     pub allow_format_lint: Option<bool>,
+    pub allow_tmai_mcp: Option<bool>,
     pub allow_patterns: Option<Vec<String>>,
 }
 
@@ -1920,6 +1922,7 @@ pub async fn get_auto_approve_settings(
         allow_fetch: aa.rules.allow_fetch,
         allow_git_readonly: aa.rules.allow_git_readonly,
         allow_format_lint: aa.rules.allow_format_lint,
+        allow_tmai_mcp: aa.rules.allow_tmai_mcp,
         allow_patterns: aa.rules.allow_patterns.clone(),
     };
 
@@ -1993,6 +1996,14 @@ pub async fn update_auto_approve_settings(
                 "auto_approve",
                 "rules",
                 "allow_format_lint",
+                toml_edit::Value::from(v),
+            );
+        }
+        if let Some(v) = rules.allow_tmai_mcp {
+            tmai_core::config::Settings::save_toml_nested_value(
+                "auto_approve",
+                "rules",
+                "allow_tmai_mcp",
                 toml_edit::Value::from(v),
             );
         }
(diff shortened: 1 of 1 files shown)


trust-delta (commented):
## Review (LGTM)

Tight, focused fix. Verified:

### Schema integrity
- `RuleSettingsResponse` (api.rs:1857) and `UpdateRuleSettingsRequest` (api.rs:1903) now mirror the `AutoApproveRules` struct in `crates/tmai-core/src/config/settings.rs:614-623` — `allow_read / allow_tests / allow_fetch / allow_git_readonly / allow_format_lint / allow_tmai_mcp / allow_patterns` is consistent across all three.
- Frontend types (`crates/tmai-app/web/src/lib/api-http.ts:570`) and the SettingsPanel toggle (`SettingsPanel.tsx:425`) already expected `allow_tmai_mcp`, confirming this was indeed a backend schema gap dropping the field on PUT.

### Persistence branch
- The new `if let Some(v) = rules.allow_tmai_mcp { save_toml_nested_value(...) }` block (api.rs:2002-2010) is structurally identical to the surrounding `allow_format_lint` / `allow_git_readonly` branches. No drift.

### Tests
- `test_update_rule_settings_accepts_allow_tmai_mcp` (api.rs:5230) directly reproduces the dropped-by-serde failure mode and would have caught the original bug.
- `test_get_auto_approve_returns_allow_tmai_mcp` (api.rs:5240) hydrates `Settings` with `allow_tmai_mcp = false` and asserts the GET handler propagates it. Together they cover both directions of the request/response round-trip without requiring a real config.toml write — appropriate scope.

Note: no test exercises the actual TOML write path (that would touch the global config file), but the existing `save_toml_nested_value` is well-trodden code and the persistence branch is a literal copy of sibling branches, so this is fine.

LGTM — would approve if not own-PR restricted.

PR #431, opened 2026-04-13
feat: PR Monitor project/author filter (exclude dependabot, other projects) (#378)

### LINKED ISSUE #378: feat: PR Monitor project/author filter (exclude dependabot, other projects)
## Problem

PR Monitor polls every **registered project** and reports every PR event. For tmai users with multiple projects configured (common in this repo: \`amitan-okusuri\`, \`tekipaki-timer\`, \`tmai\`, \`td-apps-site\`, …), the orchestrator receives PR events from all of them — including:

1. Dependabot / Renovate / bot PRs
2. PRs in repositories the user is not actively working on
3. Draft / WIP PRs
4. PRs the user didn't author (e.g., team members' PRs)

Most of these are noise for a single-developer orchestrator workflow and can overwhelm attention on genuine signals.

## Proposed design

Add per-orchestrator filters on PR Monitor before event emission:

\`\`\`toml
[orchestrator.pr_monitor_filter]
# Whitelist of project paths (absolute). Empty = all registered projects.
projects = ["/home/user/works/tmai"]

# Glob patterns of branch names to ignore. Common bot patterns pre-populated.
ignore_branches = ["dependabot/*", "renovate/*", "snyk-fix-*"]

# Author login patterns to ignore (glob or exact).
ignore_authors = ["dependabot[bot]", "renovate[bot]", "github-actions[bot]"]

# PR state filters
include_draft = false
\`\`\`

Defaults (if the section is absent): all projects, ignore common bots, exclude drafts. That matches the "probably wants only real work" intuition.

## Implementation location

\`crates/tmai-core/src/github/pr_monitor.rs\` — extend the loop that fetches PRs. Before populating \`current_state\` / emitting events, apply the filter. Rejected PRs are simply skipped (not stored, not emitted, and not seen again on next poll — which is the same effect as if they didn't exist).

## Interaction with #372 (busy-buffer)

Filtered events never reach the buffer, so they can't consume cap slots. This is important — without the filter, a flood of dependabot PRs would crowd out one real AgentStopped event that the user actually needs.

## Interaction with #376 (save bug)

Once #376 is fixed, \`[orchestrator.pr_monitor_filter]\` settings will persist. Until then, users can set the filter directly in \`~/.config/tmai/config.toml\`.

## Test plan

- Filter with matching branch pattern → PR skipped, no event
- Filter with matching author → PR skipped
- Empty / missing filter → all PRs processed (current behaviour)
- Filter mutation via hot-reload → next poll reflects new filter

trust-delta:
Closes #378.

## Summary
- Adds `pr_monitor_exclude_authors: Vec<String>` (default `["dependabot[bot]", "renovate[bot]"]`) that drops matched PRs *before* any state is recorded or CoreEvent is emitted — matching PRs never appear in `previous_states`, never auto-associate, and never fire `PrCreated`.
- Adds `pr_monitor_scope: "current_project" | "all"` (default `current_project`). In `current_project` mode, tmai skips spawning PR Monitor for registered projects whose canonical path is not an ancestor/descendant of the process cwd, so multi-project tmai instances don't flood the orchestrator with events from repos the user isn't actively working on.
- `PrInfo` now carries `author` (populated from `gh pr list --json author`). Web API request/response bodies expose the two new fields for the WebUI.

## Test plan
- [x] `cargo test -p tmai-core pr_monitor` — 23 passed (author-exclude hit/miss, empty author, empty list, warm-up baseline respecting filter, bot PRs never emitting events).
- [x] `cargo test -p tmai-core save_project_orchestrator` — 4 passed (config roundtrip covers the two new fields via `sample_non_default_orch`).
- [x] `cargo build --bin tmai` — clean build.
- [ ] Manual: with `pr_monitor_enabled = true` and a dependabot PR open, verify no `PrCreated`/`PrCiFailed` fires for that PR but a human PR in the same repo still triggers events.
- [ ] Manual: with multiple `[[project]]` entries registered and `scope = "current_project"` (default), verify only the project matching `pwd` starts a PR monitor (check `PR monitor: scope=current_project, skipping non-cwd project` in logs).

## Summary by CodeRabbit

* **新機能**
  * PR監視設定に著者除外リストと監視スコープが追加され、特定の著者を除外したり「カレントプロジェクトのみ／全プロジェクト」を切替えて監視対象を制御できます。
* **改善**
  * 監視起動時にスコープに応じたプロジェクト選別が行われ、スコープ外のプロジェクトは起動をスキップしてログに記録されます。
* **テスト**
  * 著者除外の比較・マッチングと監視振る舞いを検証する単体・統合テストを追加しました。

### DIFF
--- a/crates/tmai-app/web/src/lib/api-http.ts
+++ b/crates/tmai-app/web/src/lib/api-http.ts
@@ -447,6 +447,7 @@ export interface PrInfo {
   deletions: number;
   comments: number;
   reviews: number;
+  author?: string;
   merge_commit_sha?: string;
 }
 
@@ -644,6 +645,8 @@ export interface GuardrailsSettings {
   escalate_to_human_after: number;
 }
 
+export type PrMonitorScope = "current_project" | "all";
+
 export interface OrchestratorSettings {
   enabled: boolean;
   role: string;
@@ -653,6 +656,8 @@ export interface OrchestratorSettings {
   auto_action_templates: AutoActionTemplates;
   pr_monitor_enabled: boolean;
   pr_monitor_interval_secs: number;
+  pr_monitor_exclude_authors: string[];
+  pr_monitor_scope: PrMonitorScope;
   /** Whether this is a per-project override (true) or global fallback (false) */
   is_project_override: boolean;
 }
@@ -1066,6 +1071,8 @@ export const api = {
       auto_action_templates?: Partial<AutoActionTemplates>;
       pr_monitor_enabled?: boolean;
       pr_monitor_interval_secs?: number;
+      pr_monitor_exclude_authors?: string[];
+      pr_monitor_scope?: PrMonitorScope;
     },
     project?: string,
   ) =>
--- a/crates/tmai-app/web/src/lib/api-tauri.ts
+++ b/crates/tmai-app/web/src/lib/api-tauri.ts
@@ -14,6 +14,7 @@ import type {
   NotifySettings,
   NotifyTemplates,
   OrchestratorRules,
+  PrMonitorScope,
   SpawnRequest,
   UsageSettings,
   WorkerPermissionMode,
@@ -215,6 +216,8 @@ export const api = {
       auto_action_templates?: Partial<AutoActionTemplates>;
       pr_monitor_enabled?: boolean;
       pr_monitor_interval_secs?: number;
+      pr_monitor_exclude_authors?: string[];
+      pr_monitor_scope?: PrMonitorScope;
     },
     project?: string,
   ) => httpApi.updateOrchestratorSettings(params, project),
--- a/crates/tmai-core/src/config/mod.rs
+++ b/crates/tmai-core/src/config/mod.rs
@@ -8,5 +8,5 @@ pub use settings::{
     AuditCommand, AutoApproveSettings, CodexWsConnection, CodexWsSettings, Command, Config,
     CreateProcessSettings, EventHandling, ExfilDetectionSettings, GuardrailsSettings,
     NotifyTemplates, OrchestratorNotifySettings, OrchestratorRules, OrchestratorSettings,
-    ProjectConfig, RuleSettings, Settings, TeamSettings,
+    PrMonitorScope, ProjectConfig, RuleSettings, Settings, TeamSettings,
 };
--- a/crates/tmai-core/src/config/settings.rs
+++ b/crates/tmai-core/src/config/settings.rs
@@ -896,6 +896,31 @@ pub struct OrchestratorSettings {
     /// Polling interval for PR/CI status checks (seconds)
     #[serde(default = "default_pr_monitor_interval")]
     pub pr_monitor_interval_secs: u64,
+
+    /// PR authors whose PRs are skipped before events are emitted.
+    /// Matched exactly against `author.login` from `gh pr list`.
+    /// Default keeps dependabot/renovate noise out of the orchestrator loop.
+    #[serde(default = "default_pr_monitor_exclude_authors")]
+    pub pr_monitor_exclude_authors: Vec<String>,
+
+    /// Project scope for PR Monitor spawning. `CurrentProject` skips
+    /// registered projects whose path does not match the process cwd,
+    /// so multi-project tmai instances don't flood the orchestrator with
+    /// events from repos the user isn't actively working on.
+    #[serde(default)]
+    pub pr_monitor_scope: PrMonitorScope,
+}
+
+/// Scope filter for PR Monitor spawning.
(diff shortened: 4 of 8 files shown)


trust-delta (commented):
## Review summary

Solid filter design — exclusion correctly runs upstream of both warm-up baseline and the transition loop, so excluded PRs never pollute `previous_states` (confirmed in `pr_monitor.rs:107-154`). Good call-out in the comment at `pr_monitor.rs:116-121` that explicitly ignored PRs aren't indistinguishable from never-seen ones.

However, **the default exclude list does not match what `gh pr list --json author` actually returns on this repo**, so the defaults are a no-op today. Flagging as a blocker even though I can't formally request changes on my own PR.

### 🔴 Default `pr_monitor_exclude_authors` values don't match `gh pr list --json author` output

`crates/tmai-core/src/config/settings.rs:1278-1282` defaults to:
```rust
vec![\"dependabot[bot]\".to_string(), \"renovate[bot]\".to_string()]
```

But on this repo, `gh pr list --json author` returns:
```json
{\"author\":{\"is_bot\":true,\"login\":\"app/dependabot\"},\"number\":392}
```

i.e. gh's GraphQL-backed output uses the `app/<name>` form, not the `<name>[bot]` form that the REST API / web UI use. `is_author_excluded` is exact-match (`a == &pr.author`), so out-of-the-box dependabot PRs are **not** filtered. This is precisely the manual test item that was left unchecked in the PR body (`[ ] Manual: ... verify no PrCreated/PrCiFailed fires for that PR`).

Options:
1. Change defaults to `[\"app/dependabot\", \"app/renovate\"]` (matches gh's actual output — simplest).
2. Normalize `pr.author` and/or the compare list to strip `app/` and `[bot]` and compare the bare login (more forgiving for users copy-pasting from the web UI).
3. At minimum, land the manual test from the PR body before merging so this doesn't ship broken.

Given the whole point is \"keep dependabot noise out by default,\" shipping defaults that don't match feels like the wrong thing to merge. Please confirm with a real dependabot PR (repo has several — #390–#392) that `is_author_excluded` fires before merging.

### 🟡 `project_matches_cwd` accepts cwd-above-project, broadening scope unexpectedly

`src/main.rs:711-719`:
```rust
cwd.starts_with(proj_ref) || proj_ref.starts_with(cwd)
```

The first branch (\"I'm inside project X\") matches the stated goal. The second (\"project X is inside cwd\") means that if a user happens to launch tmai from a *parent* directory of several registered projects (e.g. `~/works` with `~/works/tmai`, `~/works/other-repo`), **all** of them spawn PR monitors — which is the \"flood the orchestrator\" scenario the PR is trying to avoid. At the extreme, `cwd=/home` would match every registered project.

Is this intentional (to support \"start from workspace root\")? If so, documenting it on the enum variant would help. If not, dropping the second branch would tighten the scope to match the stated intent.

Canonicalization handling is otherwise fine: `starts_with` is component-based so trailing slashes are a non-issue, and `..` / symlinks are covered by `canonicalize()`. One minor asymmetry — if cwd canonicalizes but project doesn't (or vice versa) you're comparing different representations and will false-negative; probably acceptable as a fallback, but worth a brief comment on `project_matches_cwd` so a future reader knows.

### 🟡 `exclude_authors` update semantics are replace-not-additive

`src/web/api.rs:1798-1800`:
```rust
pr_monitor_exclude_authors: req.pr_monitor_exclude_authors
    .unwrap_or_else(|| current.pr_monitor_exclude_authors.clone()),
```

Sending `None` preserves current (good); sending `Some(vec![\"myBot\"])` replaces wholesale, dropping the defaults. That's a reasonable design, but users who want to *add* one exclude have to first know the defaults — and those defaults are version-dependent. Suggest documenting this on `UpdateOrchestratorSettingsRequest.pr_monitor_exclude_authors` (\"replaces the entire list\") so it's explicit.

Sending `Some(vec![])` correctly disables the filter — no bug there.

### ✅ What looked right

- Filter runs before the warm-up baseline is stored (`pr_monitor.rs:122-154`) — re-enabling a previously-excluded author correctly fires `PrCreated` on the next poll because it was never in `previous_states`.
- Empty `pr.author` short-circuits `is_author_excluded` (`pr_monitor.rs:64-66`), so `list_merged_prs`' `author: String::new()` doesn't accidentally match an empty-string entry in the exclude list.
- `#[serde(default)]` on `PrInfo.author` keeps the field backward-compatible for any persisted payloads.
- `list_open_prs`' 30s cache doesn't conflict with filter edits (filter is applied at consumption time, not at cache write).
- Tests cover hit / miss / empty-author / empty-list / warm-up baseline respecting filter. Good coverage for the logic; the gap is only that no test uses the *actual* string gh emits.

PR #432, opened 2026-04-13
feat: move Agent state debounce server-side (eliminate Idle↔Processing UI noise) (#424)

### LINKED ISSUE #424: feat: move Agent state debounce server-side (eliminate Idle↔Processing UI noise)
## Problem

Agent phase/detail (Idle / Processing / Blocked / etc.) flows through a three-layer async path:

```
Poller.poll_once (500ms tick)
  → state.agents update + CoreEvent::AgentsUpdated
  → SSE onAgents → WebUI debounce → GET /api/agents
```

(`crates/tmai-core/src/monitor/poller.rs:256-350`, `crates/tmai-core/src/api/events.rs:46`, `crates/tmai-app/web/src/lib/sse-provider.tsx:25-30`)

Symptom: short Idle → Processing → Idle blips (e.g., a 500ms ToolExecution tick) reach the UI as state churn — badge flicker, list reorder noise. The poller already has a `debounce_threshold()` per transition (`crates/tmai-core/src/monitor/poller.rs:44-60`) but client also debounces on top, so transitions are "debounced twice on different scales".

## Goal

Single debounce, owned server-side, so the UI never sees sub-threshold transitions.

## Proposed direction

- Move all phase-transition debouncing into the poller / state-update path. Sub-threshold flips never become a `CoreEvent::AgentsUpdated`.
- WebUI drops the client-side debounce for state updates; renders straight from SSE.
- Snapshot endpoints (`/api/agents`) reflect post-debounce state, matching what SSE pushes.

## Open questions

- Some flips (e.g., `AwaitingApproval` arrival) should be promoted past debounce since latency is user-visible. Distinguish "interactive" vs "transient" categories.
- TUI also consumes the same state — verify it benefits identically (no separate debounce of its own).

## Related

- #422 — PR Monitor pattern (server-side SoT)
- Sibling unification issues for git / worktree

trust-delta:
## Summary

- Promote the per-subscriber agent-list fingerprint dedup from `src/web/events.rs` into `TmaiCore::notify_agents_updated()`, so sub-threshold phase flips never become a `CoreEvent::AgentsUpdated`.
- The poller already collapses Idle↔Processing oscillation via `debounce_threshold()` and overrides `agent.status` to the committed value during the pending window. Snapshot endpoints (`/api/agents`) therefore reflect post-debounce state, matching what SSE pushes.
- WebUI (`useAgents.ts`) already renders straight from SSE push — no client-side agent-state debounce existed to remove. `useWorktrees` keeps its own GET-rate-limit debounce (separate concern).
- `awaiting_approval` keeps its 0ms threshold so interactive transitions still surface immediately.

Closes #424.

## Behaviour change

| Before | After |
|---|---|
| `notify_agents_updated()` fires every 500ms regardless of state | Only fires when post-debounce fingerprint changes |
| SSE handler computes per-connection fingerprint and dedups before sending | Forwards every `AgentsUpdated` directly; trusts core-side dedup |
| Sub-threshold blip wastes broadcast slots even though SSE drops it | Sub-threshold blip never enters the broadcast channel |

## Test plan

- [x] `cargo test -p tmai-core --lib` — 1014 passed
- [x] `cargo test -p tmai --lib` — 143 passed
- [x] `cargo fmt`, `cargo clippy -p tmai-core -p tmai` clean for changed files
- [x] New unit tests:
  - `test_debounce_suppresses_sub_threshold_idle_processing_blip` — 500ms Idle→Processing→Idle blip leaves committed Idle untouched
  - `test_awaiting_approval_commits_immediately_bypassing_debounce` — approval surfaces on the very next tick
  - `test_debounce_threshold_*` — policy assertions (oscillation 500ms, approval 0ms, post-approval fast)
  - `test_notify_agents_updated_dedups_same_fingerprint` — duplicate emission suppressed
  - `test_notify_agents_updated_force_bypasses_dedup` — force-variant always emits

## Notes

- Pre-existing `cargo clippy` errors in `tmai-app` (non-exhaustive `CoreEvent` match in `crates/tmai-app/src/events.rs`) and other crates are unrelated to this PR.
- TUI consumes `PollMessage::AgentsUpdated` and calls `notify_agents_updated()` with no debounce of its own — it now benefits from the same core-side dedup automatically.

### DIFF
--- a/crates/tmai-core/src/api/core.rs
+++ b/crates/tmai-core/src/api/core.rs
@@ -57,6 +57,14 @@ pub struct TmaiCore {
     defer_registry: Arc<DeferRegistry>,
     /// Codex WebSocket sender registry for bidirectional control
     codex_ws_senders: RwLock<Option<CodexWsSenderRegistry>>,
+    /// Last agent-list fingerprint used to dedup `CoreEvent::AgentsUpdated`.
+    ///
+    /// The poller already debounces sub-threshold phase flips (see
+    /// `debounce_threshold()` in `monitor/poller.rs`), so the snapshot
+    /// returned by `list_agents()` is already post-debounce. Gating
+    /// event emission on fingerprint equality ensures subscribers never
+    /// see spurious `AgentsUpdated` notifications when nothing changed.
+    pub(crate) last_agents_fingerprint: RwLock<String>,
 }
 
 impl TmaiCore {
@@ -93,6 +101,7 @@ impl TmaiCore {
             transcript_registry,
             defer_registry,
             codex_ws_senders: RwLock::new(None),
+            last_agents_fingerprint: RwLock::new(String::new()),
         }
     }
 
--- a/crates/tmai-core/src/api/events.rs
+++ b/crates/tmai-core/src/api/events.rs
@@ -326,8 +326,57 @@ impl TmaiCore {
     ///
     /// Called by external consumers (e.g. TUI main loop) after processing
     /// `PollMessage::AgentsUpdated`. Ignored if no subscribers are listening.
+    ///
+    /// Emits `CoreEvent::AgentsUpdated` only when the post-debounce agent
+    /// snapshot actually changed since the previous emission. Because the
+    /// poller already collapses sub-threshold phase flips via
+    /// `debounce_threshold()`, this fingerprint check means sub-threshold
+    /// Idle↔Processing blips never become `AgentsUpdated` events — the UI
+    /// sees a single stable signal per real transition.
     pub fn notify_agents_updated(&self) {
+        let fingerprint = self.compute_agents_fingerprint();
+        let changed = {
+            let mut last = self.last_agents_fingerprint.write();
+            if *last == fingerprint {
+                false
+            } else {
+                *last = fingerprint;
+                true
+            }
+        };
+        if changed {
+            let _ = self.event_sender().send(CoreEvent::AgentsUpdated);
+        }
+    }
+
+    /// Force-emit `CoreEvent::AgentsUpdated` and refresh the fingerprint,
+    /// bypassing the dedup check. Used by tests and in rare cases where a
+    /// subscriber needs an unconditional refresh signal.
+    pub fn notify_agents_updated_force(&self) {
+        let fingerprint = self.compute_agents_fingerprint();
+        *self.last_agents_fingerprint.write() = fingerprint;
         let _ = self.event_sender().send(CoreEvent::AgentsUpdated);
+    }
+
+    /// Build a fingerprint over the current agent list, excluding volatile
+    /// fields (`last_update`) that change every poll regardless of phase.
+    ///
+    /// Matches the prior SSE-side dedup in `src/web/events.rs`; moving it
+    /// here means all subscribers (TUI, Tauri, MCP, SSE) share one consistent
+    /// debounced signal.
+    fn compute_agents_fingerprint(&self) -> String {
+        let agents = self.list_agents();
+        let stripped: Vec<serde_json::Value> = agents
+            .iter()
+            .filter_map(|a| {
+                let mut v = serde_json::to_value(a).ok()?;
+                if let Some(obj) = v.as_object_mut() {
+                    obj.remove("last_update");
+                }
+                Some(v)
+            })
+            .collect();
+        serde_json::to_string(&stripped).unwrap_or_default()
     }
 
     /// Notify subscribers that team data was updated.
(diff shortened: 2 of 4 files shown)


trust-delta (commented):
## Review — focus points

Concentrated on the four focus areas (fingerprint coverage, `awaiting_approval` 0ms paths, `/api/agents` ↔ SSE consistency, TUI claim).

### (1) Fingerprint coverage — acceptable, but wider than "phase only"
`compute_agents_fingerprint` at `crates/tmai-core/src/api/events.rs:362-379` only strips `last_update`. `AgentSnapshot` (`crates/tmai-core/src/api/types.rs:164-309`) still serializes:

- `cursor_x` / `cursor_y` (`#[serde(skip_serializing_if = "Option::is_none")]`)
- statusline-hook fields: `cost_usd`, `duration_ms`, `lines_added`, `lines_removed`, `context_used_pct`, `context_window_size`
- `context_warning`, `active_subagents`, `compaction_count`, `detection_source`, `detection_reason`

So any statusline hook tick or cursor movement busts the fingerprint and fires `AgentsUpdated`, even with no phase change. This is **identical to the pre-PR behavior** (same fingerprint, just moved from SSE-side to core-side), and the UI does want live cost/line/duration updates, so this is fine. Worth noting in case someone later assumes "fingerprint == phase dedup" — it isn't.

### (2) `awaiting_approval` debounce paths
Verified `debounce_threshold` at `crates/tmai-core/src/monitor/poller.rs:45-60`:

- `* → awaiting_approval` = 0ms ✔ covered by `test_awaiting_approval_commits_immediately_bypassing_debounce`.
- `awaiting_approval → processing` = 200ms ✔ covered by `test_debounce_threshold_post_approval_fast`.
- **`awaiting_approval → idle` = 200ms but not asserted in any test.** The review comment explicitly asked about the post-approval-return-to-Idle case. It works today (`from == "awaiting_approval"` branch matches both destinations at 200ms, well under 500ms oscillation threshold), but a one-line addition would pin the behavior:
  ```rust
  assert!(debounce_threshold("awaiting_approval", "idle") < Duration::from_millis(500));
  ```

### (3) `/api/agents` snapshot ↔ SSE consistency
Confirmed post-debounce semantics via `crates/tmai-core/src/monitor/poller.rs:1627-1639` and `:1651-1665`: when a transition is pending, the poller overrides `agent.status` back to the committed status before `apply_agents_poll_update` stores the agent. `list_agents()` therefore returns post-debounce state, and both `/api/agents` (api.rs:373-380) and SSE (`events.rs`) pull from the same `core.list_agents()` — they cannot diverge on phase.

Minor nit (pre-existing, not this PR): the override reconstructs `Processing { activity: Thinking }` from the coarse name rather than reusing `committed.full_status`, even though `full_status` is stored on `CommittedAgentState`. Other branch at `:1614` does use `committed.full_status.clone()`. Activity detail can briefly drop to `Thinking` during a pending window. Not regressed by this PR.

Race: SSE subscriber connects → sends initial snapshot A via `build_agents_json`, does NOT touch the shared `last_agents_fingerprint`. Next poller tick computes fingerprint, compares to the core-side `last_agents_fingerprint` (seeded `String::new()` at startup), emits `AgentsUpdated` → SSE re-sends snapshot. Potentially one redundant initial push per SSE connect if the core hasn't yet emitted, harmless.

### (4) TUI claim — the PR body overstates this
PR body says:
> TUI consumes `PollMessage::AgentsUpdated` and calls `notify_agents_updated()` with no debounce of its own — it now benefits from the same core-side dedup automatically.

Verified in `src/ui/app.rs:375-405` and `:436-489`: the TUI ingests agents **directly from `poll_rx`** (`PollMessage::AgentsUpdated(agents)`) and re-renders from that. It subscribes to `CoreEvent` (line 227) but only handles `TeammateIdle`, `TaskCompleted`, `WorktreeSetupCompleted`, `WorktreeSetupFailed`, `PromptReady`, `AgentTargetChanged`. There is no handler for `CoreEvent::AgentsUpdated` — the fingerprint dedup added in this PR does not reach the TUI at all.

The TUI's stability comes from the **poller-side debounce** (the `agent.status` override at `:1627-1665`), which is pre-existing and now well-tested by this PR. That's the real claim. Suggest rewording the PR body to: "TUI stability comes from the poller-side debounce that this PR now locks down with tests; the new core-side fingerprint dedup benefits subscribers of `CoreEvent::AgentsUpdated` (SSE, future Tauri/MCP consumers)."

### Other notes

- `notify_agents_updated_force` (events.rs:350-355) has no production callsite — only the two added tests use it. YAGNI-ish; consider dropping it and calling the private path directly from tests, or keep it as a documented escape hatch.
- `test_notify_agents_updated_dedups_same_fingerprint` relies on a 50ms timeout to assert "no event arrived." Standard tokio pattern in this repo; accept.

### Verdict

Logic is correct on all four focus points. The poller debounce + new shared fingerprint is a clean consolidation and the regression tests are valuable. Two small follow-ups worth doing (non-blocking):
1. Add `debounce_threshold("awaiting_approval", "idle")` assertion.
2. Correct the TUI sentence in the PR body (or merge commit message) — the fingerprint dedup does not flow to the TUI; the poller-side override is what helps it.

Issue #423, opened 2026-04-13
feat: unify Git state on poller-driven SoT (eliminate /api/git stateless fetch lag)


trust-delta:
## Problem

Git state for each agent's worktree is derived twice with different freshness:

- **Producer**: `Poller` runs `git rev-parse` / `git status` on each agent's cwd every ~20s and caches `GitInfo` with 10s TTL (`crates/tmai-core/src/git/mod.rs:44-56`, `crates/tmai-core/src/monitor/poller.rs:348-360`). The result populates `agent.git_branch` / `agent.git_dirty`.
- **WebUI**: `/api/git/branches`, `/api/git/diff-stat`, `/api/git/log`, `/api/git/graph` each shell out to `git` per request (`crates/tmai-app/web/src/lib/api-http.ts:877-894`).

Result: `agent.git_branch` may be stale while `/api/git/log` returns fresh output (or vice-versa). UI shows transiently inconsistent state.

## Goal

Poller becomes the single producer of git state per worktree. WebUI consumes the same in-memory model and gets pushed updates via SSE.

## Proposed direction

```
Poller git scan → tmai_core::git::cache → CoreEvent::GitStateChanged
                                          ├→ WebUI SSE (refresh)
                                          └→ /api/git/* serve from cache
```

- `/api/git/*` endpoints serve from cache; only fall through to fresh `git` calls when cache is missing on cold start.
- Poller emits `GitStateChanged` events on transitions; WebUI re-renders without polling.
- User-action driven refreshes (focus, manual refresh button) bump the poller's tick rather than running parallel git calls.

## Out of scope

- `git diff` for arbitrary commits (read-on-demand, not state-tracked) stays as-is.

## Related

- #422 — same pattern for PR Monitor (this is its sibling for git domain)

PR #433, opened 2026-04-13
feat: unify PR Monitor and WebUI on a single source of truth (CoreEvent stream) (#422)

### LINKED ISSUE #422: feat: unify PR Monitor and WebUI on a single source of truth (CoreEvent stream)
## Problem

PR Monitor and WebUI poll GitHub independently. The WebUI updates faster than PR Monitor in practice, so users see a PR's state change in the UI, then receive the matching `[PR Monitor] PR #N ...` notification several seconds later. The lag is most visible right after a `gh pr merge` or CI completion: UI flips to merged/passed first, notification arrives afterwards (and #399's typing grace adds further delay on top).

Observed during 2026-04-13 dogfooding session.

## Goal

PR Monitor becomes the single producer of PR/CI state changes. The WebUI consumes the same events. Both surfaces show transitions in lockstep, ordered by the producer's poll tick rather than two independent timers.

## Proposed direction

```
gh poll → tmai_core::pr_monitor → CoreEvent (PrCreated / PrCiPassed / PrClosed / PrCiFailed)
                                  ├→ OrchestratorNotifier (text injection)
                                  └→ WebUI SSE  (badge / list refresh)
```

- Move PR/CI state derivation out of the WebUI's polling endpoints; serve from the same in-memory model PR Monitor maintains.
- WebUI continues to fetch initial snapshot on open; subsequent updates arrive as events.
- The 30s TTL `gh` cache (`crates/tmai-core/src/github/mod.rs`) becomes irrelevant for live state — PR Monitor owns freshness.
- Asymmetric direction (UI driving PR Monitor) was considered and rejected: PR Monitor must work when no browser is open.

## Open questions

- Migration path for endpoints currently calling `gh` directly (`/api/github/prs`, `/api/github/checks`, etc.). Either re-route them through the new model, or keep them as fallback for cold-start before the first poll completes.
- PR Monitor poll interval is currently configured per-project; UI may want to drive a refresh on user action (focus, manual refresh button) — does that bump PR Monitor's tick, or stay independent?
- Coexistence with #378 (project/author filter) and #383 (default-enable) — those become non-blocking once this lands.

## Related

- #399 — typing-grace adds further perceived lag (cumulative with this issue)
- #378 — project/author filter (independent)
- #383 — default pr_monitor_enabled with orchestrator (independent)
- AutoActionExecutor (#364) — also consumes these events; benefits from a single producer

## Priority

Medium — UX polish, not a correctness issue. Good follow-up to #399 since both deal with notification timing perception.

trust-delta:
## Summary

- Make `PrMonitor` the single producer of PR/CI state changes. After every poll it publishes an in-memory `GithubSnapshot` (open PRs keyed by branch) **before** emitting `CoreEvent::Pr*`, so a listener that refetches on event always sees the post-transition state.
- `/api/github/prs` serves from that snapshot once warmed up, falling back to `gh` only on cold start. Merged PRs still go through `gh` directly (low volatility, local-branch scoped).
- `SnapshotGuard` clears the snapshot entry when the monitor task exits (normal, cancel, or panic), so endpoints don't keep serving a frozen view from a dead monitor.
- WebUI subscribes to `pr_created / pr_ci_passed / pr_ci_failed / pr_review_feedback / pr_closed` via the shared SSE provider; `BranchGraph` refetches `listPrs` on event and on SSE reconnect (laptop sleep, network blip — EventSource doesn't replay named events). The 30s PR poll is removed; issues keep a 30s timer since they have no monitor yet.

## Review fixes (post-initial)

- **snapshot publish reorder** — moved `publish_snapshot` ahead of the per-PR emit loop so the SoT contract holds across the event boundary.
- **SSE reconnect refetch** — `subscribeSSE` now exposes `onReconnect` (fires on every reopen after the first), threaded through `SSEProvider`; `BranchGraph` uses it to resync PRs.
- **monitor-exit snapshot cleanup** — `SnapshotGuard` in `spawn_pr_monitor` removes the entry when the task drops.

Closes #422.

## Test plan

- [x] `cargo test -p tmai-core` (1017 tests — 4 new: snapshot read/write/clear + warmup gating).
- [x] `cargo clippy -p tmai-core -p tmai` clean.
- [x] WebUI: `tsc -b`, `vitest run` (104 tests), `biome check`, `vite build`.
- [ ] Live dogfood: enable `pr_monitor_enabled`, open WebUI next to orchestrator terminal, trigger `gh pr merge` — confirm the badge and `[PR Monitor]` text advance on the same tick; close laptop briefly and confirm the PR list resyncs on reopen.


trust-delta (commented):
Focused review on the five points called out for #422.

### 1. Snapshot write / event broadcast ordering — needs a flip

`PrMonitor::poll` broadcasts `CoreEvent::Pr*` from inside the per-PR loop (via `emit_event`) and only calls `publish_snapshot` **after** the loop (`crates/tmai-core/src/github/pr_monitor.rs:260`). The SoT contract this PR sets up is "UI and notifications observe the same tick", but the implementation emits first and writes second. If a client's SSE round-trip ever beats the `RwLock::write().await` in `publish_snapshot`, `/api/github/prs` serves the previous tick's snapshot and the UI transiently shows stale data for an event it just received — the exact failure mode the PR is trying to eliminate.

In practice an in-process RwLock almost always wins against a browser round-trip, so this is latent rather than observable today. But the ordering is free to fix: call `publish_snapshot(&prs).await` before the disappeared/emit blocks (or at minimum before the open-PR emit loop). That makes the invariant "any client reacting to event E sees a snapshot ≥ E" unconditional instead of probabilistic.

### 2. Cold start / warm-up fallback — OK, one gap

`snapshot_for` gates on `warmed_up`, and `list_prs` falls back to `gh` on `None`. Cold start and the first poll are both covered. One unhandled path: if the user toggles `pr_monitor_enabled` off after warmup, `MONITOR_SNAPSHOTS` keeps the last snapshot forever in the `LazyLock` static — endpoints will serve a frozen list until tmai restarts. Either clear the repo's entry when the monitor task ends, or let `snapshot_for` check a "monitor-alive" bit alongside `warmed_up`. Same applies if the monitor task panics — you'd want that to invalidate the snapshot, not freeze it.

### 3. Closed/merged PR removal — OK

`publish_snapshot(&prs)` only contains the current open set, so closed PRs drop from the snapshot on the same tick a `PrClosed` event fires. UI side, `refetchPrs` replaces `prMap` wholesale via `setPrMap`, so the closed entry disappears. Merged PRs still pick up `merge_commit_sha` via the unchanged `gh` path in `list_prs`. Consistent.

### 4. BranchGraph event-driven refetch — missing self-recovery

The 30s PR polling timer is gone; only issues retain it. `BranchGraph` now relies entirely on SSE events to trigger `refetchPrs`. `EventSource` auto-reconnects, but any `PrCreated`/`PrClosed`/`PrCi*` emitted **during** a disconnect are not replayed on reconnect — the server has no SSE resume/last-event-id path for these. Result: a transient network blip, a laptop sleep/wake, or a tmai restart can leave the PR panel stale indefinitely until the next event fires. The issue timer masks this for issues but not PRs.

Cheapest fix: a single refetch on SSE (re)connect (fire `refetchPrs` once when the provider opens a new `EventSource`). Alternative: keep a very low-frequency (e.g. 5 min) PR fallback poll. Either eliminates the "silent stale" failure mode at a fraction of the old 30s cost.

### 5. Decision record — missing from the PR

The PR body points to `.claude/decisions/2026-04-13-pr-monitor-sot.md` as the template for #423/#425, but the file isn't in the diff (`git ls-tree -r pr-433 | grep -i decision` returns nothing; `.claude/decisions/` doesn't exist on the branch). Either the file wasn't committed, or it was intentionally kept local — but the PR body advertises it as reusable downstream, which it can't be without being in-tree. Please either add it or drop the reference.

### Misc

- `GithubSnapshot::open_prs` is keyed by head branch, but the `gh` fallback in `list_prs` and the merged-PR augmentation are also keyed by branch, so that's consistent.
- The warm-up baseline now also publishes a snapshot (good — closes the "first tick UI is empty" gap).
- Tests for `snapshot_for` cold/warm/publish are tight; nice.

### Verdict

Requesting changes on (5) — the decision record is load-bearing for the broader rollout story. Also strongly recommend fixing (1) ordering and (4) self-recovery before this ships; both are single-digit-line changes that convert probabilistic correctness into guaranteed correctness.

PR #441, opened 2026-04-13
feat: unify Git state on poller-driven SoT (eliminate /api/git stateless fetch lag) (#423)

### LINKED ISSUE #423: feat: unify Git state on poller-driven SoT (eliminate /api/git stateless fetch lag)

trust-delta:
## Summary
- New `GitMonitor` (`crates/tmai-core/src/git/monitor.rs`) is the single producer of per-repo branch list + commit graph state; `/api/git/branches` and `/api/git/graph` serve from its in-memory snapshot and fall back to live `git` only on cold start.
- Emits `CoreEvent::GitStateChanged { repo }` (SSE: `git_state_changed`) after publishing the snapshot so WebUI listeners observe post-transition state (same ordering as #433).
- `BranchGraph` refetches branches + graph on `git_state_changed` and on SSE reconnect, self-recovering from missed events.

## Design notes
- Mirrors the PR Monitor SoT pattern established in #433: `snapshot_for` / `clear_snapshot_for` / `SnapshotGuard` / `warmed_up` gating.
- Fingerprint covers: branch list, HEAD, per-branch last-commit time (local + remote-only), top graph SHA, total commit count. Cheap transition detection without full snapshot diff.
- `SnapshotGuard` clears the entry on task exit (panic/cancellation/shutdown) so a crashed monitor cannot leave a frozen view wedged in memory.
- `git_log` and `git_diff_stat` endpoints remain on-demand (not state-tracked) — they are per-(base, branch) queries not part of the UI's transition-sensitive path. `git diff` for arbitrary commits stays as-is per the issue's out-of-scope list.

## Test plan
- [x] `cargo test -p tmai-core --lib git::monitor` (8 tests, all passing — snapshot gating, warm-up baseline, publish, fingerprint transitions)
- [x] `cargo test --workspace --exclude tmai-app --lib` (1035 tests passing)
- [x] `cargo clippy --workspace --exclude tmai-app --all-targets` (no new warnings in changed files)
- [x] `cargo fmt --all -- --check`
- [x] `pnpm run build` (web bundle rebuilt)
- [x] `pnpm run lint` (biome clean)
- [ ] Manual: verify `/api/git/branches` and `/api/git/graph` return cached data after first poll; confirm `git_state_changed` SSE event fires when a local commit advances HEAD.

Closes #423.

## Summary by CodeRabbit

# リリースノート

## 新機能

* **Git状態監視機能を追加** - バックグラウンドで定期的にリポジトリのブランチ情報とコミットグラフを監視し、キャッシュに保存。APIレスポンスを高速化
* **イベントドリブンな状態更新** - Git状態変更を検出時に自動で通知し、リアルタイムデータの同期を改善

## チューニング

* **WebUI API性能向上** - キャッシュされたgit情報から直接提供することで、APIレスポンスを最適化


trust-delta:
## Orchestrator review — LGTM

Faithful mirror of the PR Monitor SoT pattern (#433). Ordering (snapshot before emit), `SnapshotGuard` via `Drop`, warm-up baseline not emitting, BranchGraph event+reconnect refetch — all present. Fingerprint design (branch list + per-branch last-commit time + remote-only branches + top SHA + total_count) is defensive and cheap. Approved to merge.

A few non-blocking observations worth filing as small follow-ups if they bite later:

### 1. No feature flag on spawn
`src/main.rs:629` spawns `spawn_git_monitor` unconditionally for every registered project, with a comment stating that's intentional. For a user with many registered projects (10+) this is 10 monitors * 4 polls/minute = 40 git processes/min. Not pathological for local ops, but "unconditional" deserves a settings toggle for a future owner who cares. Consider opening a follow-up if we see cost issues in dogfooding.

### 2. `DEFAULT_GRAPH_LIMIT` coupling
`GitMonitor`'s `DEFAULT_GRAPH_LIMIT = 200` (`crates/tmai-core/src/git/monitor.rs:25`) must stay in sync with the WebUI `BranchGraph`'s initial `graphLimit`. They're hardcoded in two places. The `graph.commits.len() >= params.limit || graph.commits.len() >= graph.total_count` guard in `git_graph` (src/web/api.rs:2842) handles most drift, but if the UI default ever climbs above 200 the cache is bypassed for the common case. One small touch: export the constant and consume it from the frontend types, or add a comment at both sites linking them.

### 3. Fingerprint remote-branch key
`format!("remote:{}", rb)` at `crates/tmai-core/src/git/monitor.rs:158`. If a local branch is literally named `remote:something`, it collides with a remote-only branch. Exotic but avoidable with a tuple key `(Kind, name)` where `Kind` is `Local | Remote` — zero collision surface.

### 4. Fingerprint transition tests use manual manipulation
8 tests cover warm-up gate, publish, clear, fingerprint shape. None exercise the real `poll()` path against an actual git repo. A single test that spins up a `tempdir`, runs `git init`, commits, then advances HEAD and checks `GitStateChanged` fires would pin the event-emission contract end-to-end. The current tests establish the right pieces; an integration test would lock them together.

### 5. `graph_limit` not customizable
`GitMonitor::new` hardcodes `DEFAULT_GRAPH_LIMIT`; there's no setter and `spawn_git_monitor` can't override. Future-proofing at negligible cost: `spawn_git_monitor` could accept an optional `graph_limit` (Some → override, None → default).

None of the above block this PR. Item 3 and 4 are cheapest to address now; 1, 2, 5 can be follow-ups.

Ordering / cleanup / self-recovery — all verified against #433 contract. Ship it.

PR #445, opened 2026-04-14
feat: unify Worktree state on poller-driven SoT (#425)

### LINKED ISSUE #425: feat: unify Worktree state on poller-driven SoT
## Problem

Worktree info has two independent sources:

- **Producer**: `Poller` runs `git worktree list` every ~60s (`crates/tmai-core/src/monitor/poller.rs:2009-2100 scan_worktrees`) and stores in `state.worktree_info` (`crates/tmai-core/src/state/store.rs:511`).
- **WebUI**: `/api/worktrees` serializes that state directly. Hooks (`crates/tmai-app/web/src/hooks/useWorktrees.ts:30-58`) refresh on `onAgents` SSE with 250ms debounce.

There are dedicated SSE events `worktree_created` / `worktree_removed`, but the UI relies on the broader `onAgents` debounce path, so worktree deletion / creation can show transient stale lists right after the action.

## Goal

Single producer (poller) → typed `WorktreeChanged` event → both API responses and SSE updates align.

## Proposed direction

- API serves only from `state.worktree_info` (already true), but additionally **invalidate-on-write**: any tmai action that creates / deletes a worktree (`spawn_worktree`, `delete_worktree`, `dispatch_issue`, `dispatch_review`) triggers an immediate poller re-scan rather than waiting up to 60s.
- WebUI subscribes to dedicated `worktree_*` events, dropping the `onAgents`-debounced refresh path for worktree data.
- Reduce baseline scan frequency (60s is fine when scans are event-driven).

## Related

- #422 — PR Monitor domain
- (sibling: git state unification — separate issue)

trust-delta:
## Summary

- Make the Poller the sole producer of worktree state and route every worktree-mutating action through an immediate-rescan notify, so `/api/worktrees` and the `worktree_*` SSE events reflect post-action state on the next tick instead of waiting up to ~15s for the baseline scan.
- Drop the `onAgents`-debounced refresh in the WebUI hook (the path that caused transient stale renders right after create/delete) and add `onReconnect` so SSE reconnects after laptop sleep / network blips re-sync the list.
- Lower the baseline `scan_worktrees` cadence from `% 30` (~15s) to `% 240` (~120s); event-driven action paths handle freshness, baseline only catches external `git worktree add/remove` invocations.

This is the worktree-domain application of the SoT pattern established by PR Monitor (#433) and Git Monitor (#423/#441).

## Implementation

- `crates/tmai-core/src/monitor/poller.rs`: module-level `WORKTREE_RESCAN_NOTIFY` (LazyLock<Notify>) + `pub fn request_worktree_rescan()`, re-exported from `monitor`. Wrap the main sleep in `tokio::select!` so the notify wakes the loop and forces `scan_worktrees` regardless of the cadence gate.
- `crates/tmai-core/src/api/actions.rs`: trigger from `create_worktree`, `delete_worktree`, `move_to_worktree`.
- `src/web/api.rs`: trigger from `spawn_worktree` (dispatch_issue endpoint) and `perform_dispatch_review` after the worktree lands on disk.
- `crates/tmai-app/web/src/hooks/useWorktrees.ts`: drop the `onAgents` + 250ms debounce path, keep the `worktree_*` named events, add `onReconnect: refresh`.
- Web bundle regenerated via `pnpm install --frozen-lockfile && pnpm build`.

## Test plan

- [x] `cargo test -p tmai-core --lib monitor::poller` — 21/21 pass (19 prior + 2 new)
  - `test_request_worktree_rescan_delivers_notification` — waiter parked before notify wakes
  - `test_request_worktree_rescan_stores_permit_when_no_waiter` — permit persists for the next `.notified()` (action issued during a busy `poll_once`)
  - both serialized via tests-only mutex against the shared static
- [x] `cargo test -p tmai-core --lib` — 1029/1029 pass
- [x] `cargo test -p tmai --lib` — 143/143 pass
- [x] `cargo clippy --lib --bins -- -D warnings` clean for both crates (test-only lints elsewhere are pre-existing)
- [x] `pnpm install --frozen-lockfile && pnpm build` succeeds; bundle drift check should pass since assets are committed
- [ ] Manual: dispatch_issue → worktree appears in `/api/worktrees` within ~1s (was up to ~15s)
- [ ] Manual: delete_worktree → row disappears within ~1s
- [ ] Manual: simulate SSE disconnect (DevTools offline → online) → `onReconnect` triggers `refresh()`, list re-syncs

Closes #425

## リリースノート

* **New Features**
  * Worktreeの作成、削除、移動操作後に、APIが即座に更新されるようになりました。

* **Performance Improvements**
  * Worktree一覧の更新がより迅速になり、ユーザーインターフェースの応答性が向上しました。サーバーサイドのイベント駆動スキャンにより、手動ポーリング待機時間が短縮されます。


trust-delta:
## Orchestrator review — LGTM

Clean execution of the worktree-domain SoT pattern (#433 / #423 sibling). The core primitive choice (`tokio::sync::Notify` + `notify_one`) is the right one for "rescan soon" — it coalesces bursts into a single rescan (multiple actions in the same tick → one scan that covers all of them) and both the waiter-present and no-waiter paths are exercised by tests.

### Verified against the focus points

**Immediate rescan on action path** — `crate::monitor::request_worktree_rescan()` fires from all 5 mutating sites: `create_worktree` / `delete_worktree` / `move_to_worktree` in `api/actions.rs`, plus `spawn_worktree` and `perform_dispatch_review` in `src/web/api.rs`. No site missed.

**Baseline cadence correctly de-prioritised** — `% 30 → % 240` drops the baseline to ~120s. Comment at `poller.rs:186-198` documents the intent ("only catches external `git worktree add/remove`"). Correct trade-off now that the fast path owns freshness.

**`force_worktree_scan` gate** — `force_worktree_scan || self.poll_count % 240 == 5`. Notify wakes the loop early AND bypasses the cadence gate. Consistent. One small observation (non-blocking): a notify wake also runs the full `poll_once()` agent poll, not just the worktree scan. Probably fine (agents stay current too); if it ever shows up as unnecessary work in a profile, a targeted "worktree-only-scan" fast path would be an easy follow-up.

**WebUI pivot** — `onAgents`-debounced path removed (the actual cause of transient stale renders); named `worktree_*` events kept; `onReconnect: refresh` added for laptop-sleep / network-blip recovery. Matches #433's onReconnect pattern.

**Test correctness** — the `WORKTREE_RESCAN_TEST_MUTEX` + `drain_worktree_rescan_permit()` pattern for serializing module-static-touching tests is correct. Each test's drain runs under the lock before the assertion, so permits can't leak between tests. Both the waiter-present and no-waiter-yet-permit-stored paths are covered by the two new tests.

### Non-blocking observations

1. **No end-to-end wiring test** — the primitive is exhaustively tested, but there's no single test that drives `create_worktree` action → asserts the rescan notify fires → asserts the event reaches a SSE subscriber. The wiring is literally 1 line per action site, so the risk is low, but a follow-up integration test would lock the call graph.

2. **External `git worktree` invocations still take ~120s** to surface, per the comment. Acceptable by design. If someone runs `git worktree add` in a terminal adjacent to tmai and expects it to show up instantly, they'll be surprised — worth documenting in user-facing docs but not here.

3. **`move_to_worktree` triggers rescan but only at the top-level** (before setup_commands spawn). Any worktree state change the setup script performs in-flight still waits for the baseline. Edge case.

Ordering / cleanup / self-recovery / test rigor — all present. Ship it.

Issue #447, opened 2026-04-14
refactor: extract GhClient trait to isolate gh CLI dependency


trust-delta:
## Background

`crates/tmai-core/src/github/` currently shells out to the `gh` CLI directly via `std::process::Command`. This works but creates several pain points:

- **Hard to test**: integration tests have to either install/mock `gh` or skip GitHub paths entirely
- **Error handling is scattered**: every call site re-parses `gh` stderr to decide what went wrong (see `feedback_mcp_tool_workarounds.md` for the recurring `merge_pr` / `kill_agent` failure modes)
- **Version-skew exposure**: when `gh` changes its JSON output shape, breakage shows up at runtime in unrelated features
- **CLI-not-installed UX**: error messages bubble up raw, instead of a single clear "install gh" guidance path

`tmux` is already abstracted via `RuntimeAdapter` (`TmuxAdapter` / `StandaloneAdapter`). `gh` is the remaining external CLI without that boundary.

## Proposal

Extract a `GhClient` trait inside `crates/tmai-core/src/github/`:

```rust
#[async_trait]
trait GhClient: Send + Sync {
    async fn list_prs(&self, repo: &RepoRef) -> Result<Vec<PrSummary>>;
    async fn get_pr(&self, repo: &RepoRef, number: u64) -> Result<PullRequest>;
    async fn merge_pr(&self, repo: &RepoRef, number: u64, opts: MergeOpts) -> Result<()>;
    // ...etc
}

struct GhCliClient { /* current shell-out impl */ }
struct MockGhClient { /* for tests */ }
```

Existing 30s TTL cache stays at the same layer (wraps the trait, not specific to the impl).

## Explicitly NOT in scope

- **Replacing `gh` with `octocrab`** — `gh` handles auth (`gh auth status`), keyring, GitHub Enterprise, and CI runner credentials. Re-implementing that is a larger swap with its own risks. Keep `gh` as the production impl; the trait just makes it swappable.

## Acceptance criteria

- [ ] `GhClient` trait defined, `GhCliClient` is the production impl
- [ ] At least one integration test using a `MockGhClient` for a flow currently untested (e.g., `auto_action` PR-CI-failed handling)
- [ ] Single error type with variants for: `GhNotInstalled`, `AuthExpired`, `RateLimited`, `Network`, `ParseError`, `Other(String)` — surfaced consistently to UI / MCP
- [ ] Existing call sites compile against the trait, not the concrete struct

## Why now

[...]

PR #451, opened 2026-04-15
refactor: extract GhClient trait to isolate gh CLI dependency (#447)

trust-delta:
## Summary

- Introduces a `GhClient` trait in `crates/tmai-core/src/github/client.rs` so the `gh` CLI shell-out is hidden behind a single seam. `GhCliClient` is the production impl; `MockGhClient` is a test double exposed from the crate so consumers (and the new integration test) can inject canned responses without touching the real CLI.
- Adds `GhError` (`GhNotInstalled` / `AuthExpired` / `RateLimited` / `Network` / `ParseError` / `Other`) plus a stderr classifier, so call sites get typed failure modes instead of re-parsing free-form text.
- Replaces the narrow `GithubApi` adapter in `auto_action` with `Arc<dyn GhClient>`. The existing `fetch_ci_failure_log_text` / `fetch_pr_comments_text` helpers now operate on the trait directly; unit tests use `MockGhClient`, and a new `tests/gh_client_auto_action.rs` drives `AutoActionExecutor::spawn` end-to-end through a seeded mock.

## Design notes

- Trait is dyn-compatible via `Pin<Box<dyn Future>>` — no `async-trait` dependency added.
- The 30s TTL cache stays at the `github::mod` layer (not on the trait), so it wraps any `GhClient` impl. `merge_pr`'s cache-invalidation stays in the wrapper too, keeping the trait a pure transport.
- `gh` is retained as the production backend. Replacing with `octocrab` was explicitly out of scope — `gh` handles auth (`gh auth status`), keyring, GHES, and CI runner creds.

## Acceptance criteria (#447)

- [x] `GhClient` trait defined; `GhCliClient` is the production impl.
- [x] Integration test using `MockGhClient` for the PR-CI-failed flow (`crates/tmai-core/tests/gh_client_auto_action.rs`), driving `AutoActionExecutor::spawn` across the public crate boundary.
- [x] Single `GhError` type with `GhNotInstalled`, `AuthExpired`, `RateLimited`, `Network`, `ParseError`, `Other(String)`.
- [x] Existing call sites compile against the trait — `AutoActionExecutor` now takes `Arc<dyn GhClient>` and `src/main.rs` injects `GhCliClient`.

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] `cargo test -p tmai-core --lib` — 1053 passed
- [x] `cargo test -p tmai-core --tests` — 2 passed (new integration test)
- [x] `cargo test --all-targets` — 143 bin tests pass
- [ ] Manual: exercise PR-CI-failed auto-action against a real repo after merge to confirm unchanged behavior

## Release Notes

* **Refactor**
  * GitHub クライアント層を再構築し、エラーハンドリングと操作管理を一元化しました。

* **Tests**
  * GitHub クライアント統合のための新しいテストを追加しました。

### DIFF
--- a/crates/tmai-core/src/auto_action/mod.rs
+++ b/crates/tmai-core/src/auto_action/mod.rs
@@ -14,9 +14,7 @@ pub mod templates;
 pub mod tracker;
 
 pub use resolver::{is_agent_online, resolve_target_agent, AgentRole};
-pub use service::{
-    AutoActionExecutor, GithubApi, NoopReviewDispatcher, RealGithubApi, ReviewDispatcher,
-};
+pub use service::{AutoActionExecutor, NoopReviewDispatcher, ReviewDispatcher};
 pub use templates::{render, AutoActionTemplates};
 pub use tracker::AutoActionTracker;
 
--- a/crates/tmai-core/src/auto_action/service.rs
+++ b/crates/tmai-core/src/auto_action/service.rs
@@ -10,14 +10,10 @@ use std::sync::Arc;
 use tokio::sync::broadcast;
 use tracing::{debug, info, warn};
 
-/// Pinned, boxed future returning `Option<String>` — used as the return type
-/// of `GithubApi` methods so the trait stays dyn-compatible without dragging
-/// in `async-trait` as a dependency.
-pub type GhFuture<'a> = Pin<Box<dyn Future<Output = Option<String>> + Send + 'a>>;
-
 use crate::agents::MonitoredAgent;
 use crate::api::{CoreEvent, GuardrailKind};
 use crate::config::{EventHandling, OrchestratorNotifySettings};
+use crate::github::{CiConclusion, GhClient};
 use crate::orchestrator_notify::SharedNotifySettings;
 use crate::state::SharedState;
 use crate::task_meta::{store as meta_store, SharedGuardrailsSettings};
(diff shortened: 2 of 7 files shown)


trust-delta (commented):
## Summary
Clean trait extraction. The `GhClient` / `GhError` split does exactly what the PR title claims — isolates the `gh` CLI shell-out behind a dyn-compatible seam without dragging in `async-trait`, and the 30s TTL cache intentionally stays at the outer layer so it wraps any impl. Behavior preserved 1:1 (same JSON fields, same timeouts, same cache invalidation on `merge_pr`). All 1044 tests pass locally.

The integration test at `crates/tmai-core/tests/gh_client_auto_action.rs` is a nice touch — it proves the trait seam is publicly usable (injecting `MockGhClient` into `AutoActionExecutor::spawn` from outside the crate) which wasn't possible with the old module-private `GithubApi` adapter.

## Minor findings (non-blocking)

### Dead helper
`crates/tmai-core/src/auto_action/service.rs:717` defines `mock_gh_no_comments()` but it's never called — all tests use either `MockGhClient::new()` (which returns `Err(not_seeded)` → `None` via `.ok()?`) or `mock_gh_with_single_comment(...)`. Either wire it into the \"no comments → skip\" path (e.g. the `pr_ci_passed_no_comments` test at line 1203 whose comment says `// no pr_comments → None`), or drop the helper.

### Typed errors don't reach free-function callers
`GhError::AuthExpired` / `RateLimited` are correctly classified inside `GhCliClient`, but the free-function wrappers in `github/mod.rs` (e.g. `list_open_prs`, `list_issues`, `get_pr_comments`) all do `.ok()?` and collapse every variant back to `None`. So the new variants only help new call sites that hold a `&dyn GhClient` directly — existing callers still see opaque `None`. Worth documenting on the module header, or following up with a pass that surfaces `AuthExpired` to the UI (since \"session expired\" vs \"no PRs\" is currently indistinguishable).

### `MockGhClient` ships in release builds
It's `pub` (not under `#[cfg(any(test, feature = \"test-support\"))]`), which is necessary because integration tests under `tests/` compile as a separate crate and can't reach `#[cfg(test)]` items. Intentional tradeoff, but if you ever want to strip it, gating behind a `test-support` feature would be the idiomatic path.

### `classify_stderr` permissiveness
`lower.contains(\"gh auth login\")` / `lower.contains(\"rate limit\")` are documented as intentionally loose, but they will false-positive if `gh` ever prints \"see \`gh auth login\` for details\" in an unrelated help message. Low probability, noting for posterity.

### `Option` vs required-op asymmetry in `MockGhClient`
Optional returns (`list_checks`, `get_ci_failure_log`, etc.) default to `Ok(None)` while required ones return `Err(not_seeded)`. The `mock_ci_failure_log_optional_defaults_to_none` test documents this, but it's subtle — a test that forgets to seed `list_checks` silently gets `None` instead of a loud failure. Consider making *all* unseeded ops error by default with an explicit seed-as-None helper, but not a blocker.

LGTM overall. Ship it.

trust-delta:
@coderabbitai review

PR #453, opened 2026-04-15
feat: type-sharing pipeline (utoipa + ts-rs) to eliminate Rust↔TS drift (#446)

### LINKED ISSUE #446: feat: type-sharing pipeline (utoipa + ts-rs) to eliminate Rust↔TS drift
## Background

The Web API surface (~80+ endpoints) and SSE event types are currently defined twice — once in Rust (axum handlers + `serde` types) and once by hand in TypeScript under `crates/tmai-app/web/src/`. This duplication has already produced real incidents:

- React #527 bundle drift (PRs #436 / #437 / #438)
- `feedback_parallel_pr_rebase_assets`: parallel PR rebases leave conflict markers in pre-built bundle assets

The drift will only get worse as endpoints grow.

## Proposal

Introduce a **single-source-of-truth** type pipeline:

1. **`utoipa`** — derive OpenAPI 3 schema from Rust route handlers + DTOs
2. **`ts-rs`** — derive TypeScript `.d.ts` from the same `serde` structs (including `CoreEvent` enum and SSE payloads)
3. CI gate: regenerate types and fail if `web/src/types/generated/` differs from HEAD

Scope this as a PoC first — pick one feature area (e.g., agents API + a couple of CoreEvent variants) and validate the workflow before rolling out to all 80+ endpoints.

## Out of scope

- Replacing axum routing macros wholesale
- TanStack Query migration (separate concern, see review notes)
- Runtime validation on the TS side (`zod` etc.)

## Acceptance criteria

- [ ] PoC: at least one endpoint group + one SSE event variant generated end-to-end
- [ ] CI check that fails on uncommitted regenerated types
- [ ] Documented in `CLAUDE.md` (or a decision record under `.claude/decisions/`) so contributors know not to hand-edit generated files
- [ ] Migration plan for the remaining endpoints (separate follow-up issues OK)

## Why now

Pain is **already happening** (bundle conflict markers in PRs). Doing this before contributor count grows means catching drift at PR time instead of at runtime.

trust-delta:
## Summary

Resolves #446. Introduces a single-source-of-truth type pipeline so the Web API
DTOs and SSE event shapes are defined **once** in Rust and derived downstream:

- **ts-rs** generates TypeScript into `crates/tmai-app/web/src/types/generated/*.ts`
- **utoipa** generates `generated/openapi.json` from the same serde structs
- `scripts/generate-types.sh` drives both via feature-gated `cargo test` runs
  (`ts-export` / `openapi`) and pipes output through `biome format --write`
- New CI job `types-drift` regenerates and fails the build on any uncommitted diff

## PoC coverage (per acceptance criteria)

- **Endpoint group**: `TaskMetaEntry` + `Milestone` (response body for `GET /api/task-meta`)
- **SSE event variant**: the full `CoreEvent` discriminated union (all variants),
  plus dependencies `WorktreeInfo`, `ActionOrigin`, `GuardrailKind`
- **CI gate**: `.github/workflows/ci.yml` → `types-drift` job
- **Contributor docs**: `crates/tmai-app/web/src/types/README.md` explains the pipeline,
  the "do not edit generated files" rule, and how to add new types
- **Migration plan**: documented in the README; covers 5 incremental phases
  (read-only GETs → side-effect POSTs → git/PR → remaining SSE variants → `utoipa-axum` unification)

## Opt-in, zero default-build cost

`ts-rs` and `utoipa` are behind Cargo features (`ts-export`, `openapi`) on `tmai-core`.
Ordinary `cargo build` / `cargo test` is unaffected. The pipeline only runs via
`scripts/generate-types.sh` and CI.

## Notable trade-offs

- Rust `u64` becomes TypeScript `bigint` (strictly more correct than the existing
  hand-written `number`; the README calls this out).
- Handler signatures and utoipa `#[path]` stubs are still separate sources; a
  follow-up will adopt `utoipa-axum` to unify them.
- Generated files are committed (reproducible bundles, reviewable type diffs).

## Test plan

- [ ] `cargo check` passes
- [ ] `cargo clippy -- -D warnings` passes
- [ ] `cargo fmt --check` passes
- [ ] `cargo test` (root) passes — 143 tests
- [ ] `bash scripts/generate-types.sh` produces identical output to HEAD (drift check)
- [ ] `crates/tmai-app/web` → `npx biome check src/` clean
- [ ] `crates/tmai-app/web` → `npx tsc --noEmit` clean
- [ ] `crates/tmai-app/web` → `pnpm build` succeeds, `src/web/assets/` unchanged (existing bundle drift check)
- [ ] New `types-drift` CI job green


trust-delta:
@coderabbitai review

trust-delta (commented):
## Review

Clean, well-scoped PoC. The feature gating (`ts-export` / `openapi`), drift-CI, and doc-stub approach keep the zero-cost-by-default promise intact. LGTM modulo the notes below — all non-blocking.

### Non-blocking observations

1. **`asCoreEvent` performs no runtime validation** (`crates/tmai-app/web/src/types/sse-event.ts:18-27`).
   The tag is coerced with `as CoreEventTag` without checking against the known discriminants, so a bogus server payload would be returned as a typed `CoreEvent`. Fine for a PoC, but if this becomes the canonical SSE narrower (per the migration plan), consider a generated `readonly Set<CoreEventTag>` guard so the function actually behaves like `unknown -> CoreEvent | null`.

2. **Stale comment in the same file** (`sse-event.ts:20-22`).
   The comment says "some paths omit [`type`] and rely on the event name exclusively", but with `#[serde(tag = \"type\")]` the `type` field is always present in the payload, so the `?? eventName` fallback should never trigger. Either drop the fallback or adjust the comment.

3. **`u64` → OpenAPI `integer/int64` with `minimum: 0`** (`crates/tmai-app/web/src/types/generated/openapi.json`, e.g. `issue`, `pr`).
   `int64` is semantically signed (max ~9.2e18) while `u64` is ~1.8e19. Not a practical issue for PR/issue numbers, but worth a README note alongside the existing \"u64 → bigint\" callout, since external OpenAPI consumers may codegen signed types.

4. **`npx biome format` in `scripts/generate-types.sh`** assumes `crates/tmai-app/web/node_modules` is populated. CI handles this via the preceding `pnpm install --frozen-lockfile`; locally a contributor running the script from a fresh checkout will trigger `npx`'s interactive download prompt. A short preflight check (or an explicit \"run `pnpm install` first\" note in the README) would be friendlier.

5. **`openapi.json` has no trailing newline** (diff shows `\\ No newline at end of file`). Cosmetic — biome only formats `.ts` output in the script, so the JSON file skips formatting entirely. Consider appending `\\n` in `regenerate_openapi_json` or piping through `jq .` so the file plays nicely with editors/linters.

6. **`regenerate_openapi_json` test wording** (`crates/tmai-core/src/api/openapi.rs:1130`).
   The \"running this in ordinary `cargo test` would race with ts-rs export_bindings tests\" doc comment is a bit misleading — the generator script already serializes the two `cargo test` invocations, so the race concern is moot. Consider rewording to \"opt-in so `cargo test` doesn't write into the tree\".

### What I like

- Feature-gated deps keep default builds unchanged; `ts-rs` and `utoipa` never touch normal `cargo build`.
- `types-drift` CI job is the right forcing function; the error message even tells contributors exactly what to run.
- Doc-stub pattern (`get_task_meta_doc`) is pragmatic for a PoC without pulling utoipa into the bin crate.
- README migration plan is concrete (5 phases, specific endpoint buckets), which is the hard part of rolling this out.

PR #465, opened 2026-04-15
feat: configurable worker model per agent type / role / dispatch (#435)

### LINKED ISSUE #435: feat: configurable worker model per agent type / role / dispatch
## Problem

Spawned workers (via \`dispatch_issue\`, \`dispatch_review\`, \`spawn_worktree\`, \`spawn_agent\`) inherit the user's default Claude/Codex/Gemini model. For an orchestrator running Opus 1M, every dispatched implementer also runs Opus 1M, which is overkill for small, well-specified issues (#428-style field additions, #395-style test fixes) and burns context / cost unnecessarily.

Current invocation at \`src/web/api.rs:3219\` (and the parallel \`launch_agent_in_worktree\`, \`dispatch_review\` paths) runs \`claude\` with no \`--model\` flag, so the user's global setting applies.

## Proposal — three-tier configuration

```toml
[spawn.claude]
default_model = \"claude-sonnet-4-6\"          # implementers, reviewers default
orchestrator_model = \"claude-opus-4-6\"        # optional role override

[spawn.codex]
default_model = \"o3\"
# orchestrator_model = \"o3-high\"

[spawn.gemini]
default_model = \"gemini-2.5-flash\"
# orchestrator_model = \"gemini-2.5-pro\"
```

### Resolution order (highest priority wins)

1. **Per-dispatch override** — optional \`model\` parameter on \`dispatch_issue\`, \`dispatch_review\`, \`spawn_agent\`, \`spawn_worktree\`, \`spawn_orchestrator\` (ad-hoc override per call)
2. **Role-specific setting** — \`[spawn.<type>].orchestrator_model\` (if the spawn is an orchestrator role)
3. **Type default** — \`[spawn.<type>].default_model\`
4. **User's global default** — if none set, fall through to the current no-flag behavior

Roles to distinguish:
- \`orchestrator\` (spawn_orchestrator, \"claude (main)\" dropdown)
- \`implementer\` (dispatch_issue, spawn_worktree, spawn_agent)
- \`reviewer\` (dispatch_review)

## Scope

- **Settings**: add \`[spawn.<agent_type>]\` tables with \`default_model\` / \`orchestrator_model\` / (later) \`implementer_model\` / \`reviewer_model\` if needed
- **Spawn pipeline**: resolve model at spawn time, inject \`--model <name>\` into \`claude\` args (and equivalent flags for Codex / Gemini — existing code at \`tmai wrap codex --model o3\` shows the Codex path; Gemini uses \`--model\` too)
- **MCP tools**: add optional \`model\` parameter to dispatch / spawn tools; omit from body when unset
- **WebUI**: reflect current defaults in settings panel (read-only initially is fine; editable later)

## Out of scope (follow-ups)

- Per-project override (add \`[project.spawn.<type>]\`) — matrix gets big, defer
- Automatic role detection from issue size / type — heuristic, not necessary for v1

## Test plan

- Resolution order unit tests for each of the 4 tiers
- Round-trip via settings save/load
- Integration: \`dispatch_issue\` with and without \`model\` param → correct \`--model\` in spawned command
- Codex / Gemini parity tests for the flag name

## Priority

P2 — quality-of-life for users who want to save on Opus cost/context, not a correctness issue. Aligns with the "orchestrator=Opus, implementers=Sonnet" pattern observed during dogfooding.

trust-delta:
Closes #435.

## Summary

Adds a four-tier model resolution chain so the orchestrator can run on Opus while implementers/reviewers default to a cheaper model:

1. **Per-dispatch override** — optional `model` argument on `dispatch_issue`, `dispatch_review`, `spawn_agent`, `spawn_worktree`, `spawn_orchestrator` (HTTP body **and** MCP tool param)
2. **Role-specific** — `[spawn.<type>].orchestrator_model`
3. **Type default** — `[spawn.<type>].default_model`
4. **CLI default** — falls through to no `--model` flag (current behaviour)

Subtables added: `[spawn.claude]`, `[spawn.codex]`, `[spawn.gemini]`. Other role-specific fields (`implementer_model` / `reviewer_model`) are deferred per the issue.

## Notes

- All three CLIs accept the same `--model <name>` flag, so injection uses one helper.
- Empty / whitespace strings collapse to "unset" at every layer so a user can blank a field via `PUT /api/settings/spawn` to fall back to the CLI default.
- `inject_model_flag()` is a no-op for `bash` / `sh` / `zsh` and unrecognised commands.
- `perform_dispatch_review` gains a trailing `model_override: Option<String>`. Pre-existing `tmai-app` build error in `crates/tmai-app/src/events.rs` (non-exhaustive `CoreEvent` match) is unrelated and reproduces on `main`.
- WebUI does not currently render a spawn-settings panel; the issue calls out a panel as "later, editable later" — endpoints are exposed and ready when one is added.

## Test plan

- [x] `cargo test -p tmai-core --lib config::` — 7 new resolver/round-trip tests pass
- [x] `cargo test -p tmai --lib web::api::tests::inject_model` — 5 new arg-injection tests pass
- [x] `cargo clippy -p tmai -p tmai-core --no-deps -- -D warnings` — clean
- [x] `cargo fmt --all --check` — clean
- [x] Pre-existing `tmai-app` build break confirmed unrelated (reproduces on stashed `main`)

### DIFF
--- a/crates/tmai-core/src/config/mod.rs
+++ b/crates/tmai-core/src/config/mod.rs
@@ -5,8 +5,9 @@ pub use claude_settings::{
     ClaudeSettings, ClaudeSettingsCache, SpinnerVerbsConfig, SpinnerVerbsMode,
 };
 pub use settings::{
-    AuditCommand, AutoApproveSettings, CodexWsConnection, CodexWsSettings, Command, Config,
-    CreateProcessSettings, EventHandling, ExfilDetectionSettings, GuardrailsSettings,
-    NotifyTemplates, OrchestratorNotifySettings, OrchestratorRules, OrchestratorSettings,
-    PrMonitorScope, ProjectConfig, RuleSettings, Settings, TeamSettings,
+    AgentModelSettings, AuditCommand, AutoApproveSettings, CodexWsConnection, CodexWsSettings,
+    Command, Config, CreateProcessSettings, EventHandling, ExfilDetectionSettings,
+    GuardrailsSettings, NotifyTemplates, OrchestratorNotifySettings, OrchestratorRules,
+    OrchestratorSettings, PrMonitorScope, ProjectConfig, RuleSettings, Settings, SpawnRole,
+    SpawnSettings, TeamSettings,
 };
--- a/crates/tmai-core/src/config/settings.rs
+++ b/crates/tmai-core/src/config/settings.rs
@@ -825,6 +825,94 @@ pub struct SpawnSettings {
     /// `"default"` (or any non-`claude` command) to skip injection.
     #[serde(default = "default_worker_permission_mode")]
     pub worker_permission_mode: String,
+
+    /// Per-agent-type model configuration for Claude Code workers.
+    #[serde(default)]
+    pub claude: AgentModelSettings,
+
+    /// Per-agent-type model configuration for Codex CLI workers.
+    #[serde(default)]
+    pub codex: AgentModelSettings,
+
+    /// Per-agent-type model configuration for Gemini CLI workers.
+    #[serde(default)]
+    pub gemini: AgentModelSettings,
+}
+
+/// Per-agent-type model selection settings.
+///
+/// Each field is optional. Resolution falls back through the chain:
+/// per-dispatch override → role-specific → `default_model` → no flag
+/// (CLI uses its own default).
+#[derive(Debug, Clone, Default, Serialize, Deserialize, PartialEq, Eq)]
+pub struct AgentModelSettings {
+    /// Model used when no role-specific override applies (e.g. implementer,
+    /// reviewer, or any role without a dedicated field).
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub default_model: Option<String>,
+
+    /// Model used when the spawn role is `Orchestrator`.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub orchestrator_model: Option<String>,
+}
+
+/// Role of a spawned agent — governs which field of `AgentModelSettings` is
+/// consulted after the per-dispatch override is checked.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum SpawnRole {
+    /// Orchestrator role (`spawn_orchestrator`).
+    Orchestrator,
+    /// Implementer role (`dispatch_issue`, `spawn_worktree`, `spawn_agent`).
+    Implementer,
+    /// Reviewer role (`dispatch_review`).
+    Reviewer,
+}
+
+impl SpawnSettings {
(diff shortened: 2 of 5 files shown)


trust-delta (commented):
## Review summary

Clean, well-tested implementation of the 4-tier model resolution chain. The design choices (basename-based CLI detection, whitespace-as-unset normalization, per-dispatch override propagated through both HTTP and MCP layers) all look right, and the test coverage (7 resolver tests + 5 `inject_model_flag` tests + TOML round-trip) covers the important edge cases including blank override, unknown agent types, and orchestrator-specific fallback.

Spot-checked the spawn paths — callers of `spawn_in_pty`/`spawn_in_tmux` (`launch_agent_in_worktree`, `spawn_orchestrator`, `spawn_worktree`, `perform_dispatch_review`) all set `model: None` on the inner `SpawnRequest` after injecting the flag themselves, so there's no double-injection risk.

(Cannot formally approve own PR — leaving LGTM as a comment.) A few minor observations below — none blocking.

### Nits / optional follow-ups

1. **Minor duplication of resolve-and-push logic** — `src/web/api.rs` has the same `resolve_model(...) → push --model` block inlined in 3 places (`spawn_orchestrator` ~L2640-L2648, `perform_dispatch_review` ~L3412-L3420, `spawn_worktree` ~L3610-L3618) alongside the `inject_model_flag` helper used by `spawn_agent`/`launch_agent_in_worktree`. All 5 sites could share one helper that returns `Vec<String>` or accepts an `&mut Vec<String>` with role/override. Not worth a follow-up PR on its own, but worth folding in next time this area is touched.

2. **Dead defensive check** — `inject_model_flag` (`src/web/api.rs:586-590`) checks `if !name.is_empty()` inside `if let Some(name) = resolved`, but `resolve_model` already filters whitespace/empty via its `normalize` closure (`crates/tmai-core/src/config/settings.rs:92-94`). Same pattern in the 3 inline sites. Harmless but can be simplified.

3. **Clearing writes `= \"\"` rather than removing the key** — `persist_model_setting` (`src/web/api.rs:530-548`) writes an empty string when the user clears a field via `PUT /api/settings/spawn`. The resolver handles this correctly (treats as unset), but the `config.toml` ends up with lines like `default_model = \"\"` instead of an absent key. Purely cosmetic; leaves the config slightly noisier over time.

4. **No model-name validation** — user-supplied `model` strings are passed as CLI args. Safe against shell injection since they go through `Command::args`, not a shell, and each CLI will reject unknown names itself. Just noting this is an intentional design choice, not an oversight.

5. **utoipa schemas unaffected** — confirmed `src/web/api.rs` types don't carry `ToSchema` derives (those live in `crates/tmai-core/src/api/types.rs`), so no OpenAPI regeneration needed.

LGTM.

PR #466, opened 2026-04-15
feat: origin-aware notification filtering (suppress self / split Human vs Agent) (#440)

### LINKED ISSUE #440: feat: origin-aware notification filtering (suppress self / split Human vs Agent)
## Problem

\`CoreEvent::ActionPerformed\` fires for every side-effect action (dispatch_issue, kill_agent, approve, etc.) and is surfaced to the orchestrator as a text injection. Two practical annoyances during dogfooding:

1. **Self-echo**: when the orchestrator itself performs an action (e.g. I call \`dispatch_issue\`), tmai notifies me back with \`[tmai] Agent (mcp) performed dispatch_issue: ...\`. Pure noise — the caller already knows.
2. **Human vs Agent coupling**: \`Human (webui) performed kill_agent ...\` and \`Agent (mcp) performed ...\` currently share the same notification pipeline. A user may want to see one and not the other, or route them differently.

The existing filter (\`feedback_action_notification_noise.md\`) narrowed *which actions* emit, but not *which origin*.

## Proposed

Add an \`ActionOrigin\`-aware layer on top of the existing per-event handling mode:

### Layer 1 — suppress self (default ON)

If the \`ActionPerformed\` event's origin resolves to the notifying orchestrator itself (same agent id / same pane), skip delivery. The orchestrator already knows it called the action; the echo adds no value.

Single global toggle in Settings: **"Skip notifications for actions I performed"** (default checked).

### Layer 2 — per-origin filter matrix

For the remaining (non-self) events, let the user configure per-origin visibility:

| Origin | Default |
|--------|---------|
| Human (webui / CLI) | notify |
| Agent (mcp / another orchestrator / auto_action) | notify |
| System (auto-cleanup, escalation) | notify |

UI: one row per origin under the Notifications section, each a three-state (off / notify / notify-inline) radio. Advanced users get full control; defaults preserve current behaviour.

### Layer 3 (future, not scoped here)

Per-event × per-origin matrix. Too big to design now without data on what people actually want.

## Implementation sketch

- \`CoreEvent::ActionPerformed { origin, actor_id, action, summary }\` already has enough info (per memory \`2026-04-09-policy-orchestrator-rearchitecture.md\`).
- Core-side filter checks \`origin\` and \`actor_id == notifying_agent_id\` before calling the sender.
- Settings: add \`notify.suppress_self: bool\` and \`notify.origin_modes: HashMap<Origin, Mode>\` to \`OrchestratorNotifySettings\`.
- WebUI: add a subsection \"Sources\" to the Notifications panel — 1 checkbox for self-suppress + 3 radios for Human/Agent/System.

## Test plan

- Unit: self-originated \`ActionPerformed\` with \`suppress_self = true\` is dropped; with \`false\` is delivered.
- Unit: per-origin filter routes correctly.
- Round-trip: settings save/load covers new fields.

## Priority

P2 — quality-of-life for active orchestrator use; not broken, just noisy.

## Related

- \`feedback_action_notification_noise.md\` — this issue is the next layer on top
- #364 AutoActionExecutor — auto_action origin is a first-class case for the filter
- Sibling issue: Notifications setting copy rewrite (filed separately)

trust-delta:
## Summary

- Adds an `ActionOrigin`-aware filter layer to `OrchestratorNotifier`. `ActionPerformed` events are now gated by four new `OrchestratorNotifySettings` booleans: `suppress_self`, `notify_on_human_action`, `notify_on_agent_action`, `notify_on_system_action` — all defaulting to ON so legacy configs keep current behaviour.
- WebUI Notifications panel grows a new "Sources" subsection: a self-suppress toggle plus per-origin Notify on/off for Human / Agent / System initiators. Wired through the existing `updateOrchestratorSettings` PUT and live-reloaded via the same shared `notify_settings` lock.
- Per-event handling (`EventHandling`) remains the first gate; the origin filter only narrows `ActionPerformed`.

Closes #440.

## Test plan

- [x] `cargo test -p tmai-core --lib` (1061 passed, includes 7 new origin-filter tests + TOML round-trip)
- [x] `cargo test -p tmai` (151 passed)
- [x] `cargo fmt --all` clean
- [x] `cargo clippy -p tmai-core -p tmai --tests` — no new warnings introduced
- [x] `pnpm exec tsc -b` clean, `pnpm exec biome check` clean
- [x] `pnpm build` rebuilds bundled assets under `src/web/assets/`
- [ ] Manual smoke: toggle each Source row in WebUI Settings → verify a corresponding `ActionPerformed` (e.g. webui `kill_agent`, mcp `merge_pr`, system `auto_cleanup`) is dropped/delivered as expected


trust-delta (commented):
## Review — LGTM (cannot formally approve as PR author)

Clean, focused implementation. The origin-aware filter slots neatly in front of the existing self-echo suppression and the `default_true` fields keep legacy configs behaving identically.

### What I checked
- `build_notification` match arms in `crates/tmai-core/src/orchestrator_notify/service.rs:562-588` — precedence is correct. For an `Agent { is_orchestrator: true }` the first arm's `suppress_self` guard wins; when `suppress_self=false` it falls through to the agent-filter arm, so disabling "Agent actions" still covers orchestrator self-echoes as a secondary gate. Non-matching guards correctly fall through to `_ => {}`.
- Hot-reload path at `src/web/api.rs:1850` (`*ns.write() = updated.notify`) carries the four new fields without extra wiring.
- `OrchestratorNotifySettings::default()` at `crates/tmai-core/src/config/settings.rs:1148-1151` matches the `#[serde(default = "default_true")]` attributes — legacy TOML round-trips back to all-on, and the new `test_notify_settings_origin_fields_default_on` test pins that.
- 7 new tests cover the independent-filter contract (e.g. `test_action_performed_human_origin_passes_when_agent_filtered`) and the TOML round-trip. The existing `test_action_performed_from_orchestrator_skipped` is preserved as the default-on regression.

### Minor nits (non-blocking)

- **Label ambiguity in the "Sources" subsection** (`crates/tmai-app/web/src/components/settings/SettingsPanel.tsx:1643-1665`). When `suppress_self=off`, orchestrator-initiated actions are still Agents and therefore still gated by the "Agent actions" toggle. A reader who turns `suppress_self` off expecting to see their own actions could then be surprised when "Agent actions" is also off. A one-liner like "Orchestrator actions also count as Agent actions unless Skip-self is on" under the Agent row would remove the ambiguity. Not worth a re-roll.
- **Optimistic-update rollback captures stale state** in `setOriginFlag` (`SettingsPanel.tsx:1452-1466`): the catch branch calls `setOrchestrator(orchestrator)` with the closed-over value, so two rapid toggles that both fail could revert to a stale snapshot. The rest of this file uses the same pattern, so it's consistent — just flagging as a known rough edge.

### Not checked
- Manual smoke of the three bullet scenarios (webui `kill_agent`, mcp `merge_pr`, system `auto_cleanup`) in the PR description — the test_plan leaves that unchecked. Worth confirming before merge since the wiring is straightforward but end-to-end behaviour is the actual deliverable.

PR #468, opened 2026-04-15
feat: default pr_monitor_enabled to true when orchestrator enabled, add startup guidance (#383)

### LINKED ISSUE #383: feat: default pr_monitor_enabled to true when orchestrator enabled, add startup guidance
## Problem

\`pr_monitor_enabled\` defaults to \`false\`. Without it, the PR Monitor service does not start at tmai launch, and no PR/CI events (\`PrCreated\`, \`PrCiPassed\`, \`PrCiFailed\`, \`PrReviewFeedback\`, \`PrClosed\`) are emitted. Orchestrator agents that rely on these events for reactions get silence. During dogfooding, this silently broke the feedback loop for an entire multi-hour session before being diagnosed.

Contributing factors:
- Default-off setting means new users never see it
- WebUI toggle silently reverted on save (#376) — so even when users enabled it, it didn't persist
- No startup log or UI cue that the monitor is disabled

### 1. Default to \`true\` when orchestrator is enabled

Currently \`pr_monitor_enabled: false\`. Change the default-generation path to:

\`\`\`rust
impl Default for OrchestratorSettings {
    fn default() -> Self {
        Self {
            enabled: true, // unchanged
            pr_monitor_enabled: true, // NEW default
            // …
        }
    }
}
\`\`\`

Users who never configured it get the monitor running automatically. Existing configs with explicit \`false\` still honour that.

Alternative: leave the default as-is but change the startup log level from \`debug\` to \`info\` so the "disabled" state is visible.

### 2. Startup visibility

On tmai launch, log at \`info\` level:

\`\`\`
tmai: PR Monitor disabled (orchestrator.pr_monitor_enabled = false)
tmai: PR/CI events will not fire — enable via WebUI Settings → Orchestrator, or set pr_monitor_enabled = true in ~/.config/tmai/config.toml
\`\`\`

Visible both with and without \`--debug\`.

### 3. WebUI cue

In the Orchestrator settings panel, when \`pr_monitor_enabled\` is off, show an inline warning:

\`\`\`
⚠ PR Monitor is disabled. CI-pass / PR-comment / agent-stopped events
  that rely on PR state polling will not reach the orchestrator.
\`\`\`

### 4. Depends on #376

The default change is pointless until #376 lands — otherwise toggling the WebUI value still reverts after restart. Merge order: #376 → this issue.

## Scope boundaries

- Do NOT change \`pr_monitor_interval_secs\` default (60s is fine)
- Do NOT force-enable for users with explicit \`pr_monitor_enabled = false\`
- Do NOT add pre-existing-PR flood suppression here (that's #377's scope)

## Tests

- \`OrchestratorSettings::default().pr_monitor_enabled == true\`
- Existing config TOML with no \`pr_monitor_enabled\` key loads with default \`true\`
- Existing config with \`pr_monitor_enabled = false\` loads as \`false\` (no override)
- Startup log emitted at appropriate level

## Related

- #376 (save bug) — **blocker**
- #377 (cold-start flood) — becomes more visible if this defaults to on; fix first or land together
- Session dogfooding (2026-04-12): missing notifications traced back to this

## Priority

Low-medium. Cosmetic default change, but prevents invisible misconfiguration in practice.

trust-delta:
## Summary

- Flip `OrchestratorSettings::pr_monitor_enabled` default from `false` to `true` (struct default + `serde(default = "default_true")`) so a freshly-installed tmai with the orchestrator turned on receives PR/CI events without an extra opt-in step. Configs that explicitly set `pr_monitor_enabled = false` continue to honour the value.
- Log at `tracing::info` on tmai launch when the monitor is disabled, with guidance pointing at the WebUI toggle and the `~/.config/tmai/config.toml` key.
- Show an inline amber warning in the Orchestrator settings panel when the toggle is off.

Default-off was the silent failure mode behind a multi-hour dogfooding session where the orchestrator never received `PrCreated` / `PrCiPassed` / `PrCiFailed` / `PrReviewFeedback` / `PrClosed` events. #376 (save bug) is already merged, so the toggled value now persists across restarts.

Closes #383.

## Test plan

- [x] `cargo test -p tmai-core --lib` — 1081 tests pass, including 4 new `pr_monitor_enabled_*` cases:
  - `pr_monitor_enabled_defaults_to_true_when_struct_constructed`
  - `pr_monitor_enabled_missing_key_loads_as_true` (TOML without the key picks up `default_true`)
  - `pr_monitor_enabled_explicit_false_is_honoured` (no override of explicit user opt-out)
  - `pr_monitor_enabled_explicit_true_is_honoured`
- [x] `cargo fmt --check` clean
- [x] `cargo build` workspace-wide clean
- [x] Frontend `pnpm run build` + `pnpm run lint` clean; bundled assets refreshed
- [ ] Manual smoke: launch tmai with `pr_monitor_enabled = false` in config, confirm the two `info` lines appear in the log
- [ ] Manual smoke: open WebUI Orchestrator settings, toggle PR Monitor off, confirm amber warning appears


trust-delta (commented):
## Review (comment — cannot approve own PR)

Clean, well-scoped fix for the silent default-off issue. LGTM with minor nits.

### Correctness
- `#[serde(default = "default_true")]` + updating `OrchestratorSettings::default()` in `crates/tmai-core/src/config/settings.rs:983` / `:1393` is the right combination — serde `default` only triggers on a missing key, so explicit `pr_monitor_enabled = false` is preserved. The four new tests (`pr_monitor_enabled_defaults_to_true_when_struct_constructed`, `…missing_key_loads_as_true`, `…explicit_false_is_honoured`, `…explicit_true_is_honoured`) cover exactly the paths that matter. The existing `test_settings_defaults` in `pr_monitor.rs:1118` was correctly flipped.
- The UX surfaces (startup `tracing::info` pair in `src/main.rs:590-599` and amber banner in `SettingsPanel.tsx:1833-1842`) land the "make misconfiguration loud" goal without being noisy on the happy path.

### Minor observations (non-blocking)

1. **Warning vs orchestrator-disabled** — `src/main.rs:590`: the log fires whenever `pr_monitor_enabled == false`, even when `orch_settings.enabled == false` (i.e. the orchestrator isn't running at all). A user who deliberately runs without the orchestrator will still see "PR/CI events will not fire" on every launch. Consider also gating on `orch_settings.enabled`.

2. **Workspace vs project scope** — the warning uses `resolve_orchestrator(None)` (workspace-level), but the actual start loop in `src/main.rs:615` re-checks per-project (`project_orch.pr_monitor_enabled`). If a user relies on per-project overrides (workspace `true`, one project `false` or vice versa), the startup log will be out of sync with what's actually running. Low priority — current callers almost certainly use workspace-level only.

3. **Log level** — `tracing::info!` for a misconfiguration notice is fine given the `tmai=debug,tmai_core=debug` default, but `tracing::warn!` would be more conventional and survives stricter downstream filters. Your call.

### Nits
- Amber banner copy matches the log wording — nice consistency across UI/log.

Issue #334, opened 2026-04-10
feat: warn or auto-rebase when spawning branch from stale main


trust-delta:
## Problem

When creating a feature branch from a local `main` that is behind `origin/main`, the resulting PR has merge conflicts from the start and CI is silently skipped. This was observed when `feat/agent-metadata-semantic-addressing` was branched from a stale main — PRs #329 and #330 had already been merged, causing conflicts in `queries.rs` and `pr_monitor.rs` that blocked CI entirely without any visible error.

## Proposed Solution

Add a check at branch creation time (in `spawn_worktree` / `dispatch_issue` flow) that:

1. **Detects staleness**: Compare local base branch HEAD with `origin` before creating the worktree
2. **Auto-fetch + rebase**: `git fetch origin && git rebase origin/main` before branching, or at minimum warn the orchestrator
3. **Post-PR check**: If a PR is created and CI doesn't start within N minutes, surface the issue (conflicts, Actions disabled, etc.)

## Context

Discovered during dogfooding (2026-04-10). The orchestrator created a branch, pushed, and created a PR, but CI never ran. The root cause (merge conflicts) was only visible on the GitHub web UI — neither `gh pr checks` nor tmai surfaced the problem until manually investigated.

trust-delta:
Moved to https://github.com/trust-delta/tmai-core/issues/24 as part of the repo split. Closing here to consolidate follow-up.

PR #469, opened 2026-04-15
feat: warn or auto-rebase when spawning branch from stale main (#334)

trust-delta:
## Summary
- `spawn_worktree` / `dispatch_issue` now opt into auto-fetching `origin/<base>` and bypass a stale local base by branching off the upstream ref directly, so new PRs no longer open with merge conflicts that silently kill CI (#334).
- Surfaces a `CoreEvent::ActionPerformed { action: "spawn_worktree_stale_base", ... }` notification whenever the bypass kicks in, so the orchestrator hears about it instead of having to reverse-engineer "unexpected" commits on the new branch.
- New `WorktreeCreateRequest::auto_fetch_base` is opt-in; `dispatch_review` (already pins `origin/<base>`) and the TUI manual-create path keep it off to preserve current semantics.

## Test plan
- [x] `cargo test -p tmai-core --lib worktree::` — 30 tests pass, including 4 new staleness scenarios (stale base uses origin, up-to-date is no-op, no-origin falls back silently, opt-out skips the check).
- [x] `cargo test -p tmai --lib` — 156 tests pass, no regressions in the web spawn handler.
- [x] `cargo check -p tmai-core -p tmai --tests` clean.
- [x] `cargo fmt` / clippy clean for the changed code paths.

## リリースノート

* **New Features**
  * ワークツリー作成時にベースブランチの遅延を自動検出し、ローカルブランチがリモートに遅れている場合はリモートブランチを使用するようになりました。
  * 遅延検出時に詳細情報（ブランチ名、コミット差分数、使用したRef）がレポートされるようになりました。

* **Chores**
  * ワークツリー作成時の処理を最適化し、冗長なフェッチを削減しました。

### DIFF
--- a/crates/tmai-core/src/worktree/mod.rs
+++ b/crates/tmai-core/src/worktree/mod.rs
@@ -10,5 +10,6 @@ pub use ops::{
     check_worktree_clean, create_worktree, delete_worktree, move_to_worktree, run_setup_commands,
 };
 pub use types::{
-    WorktreeCreateRequest, WorktreeDeleteRequest, WorktreeMoveRequest, WorktreeOpsError,
+    BaseStalenessReport, WorktreeCreateRequest, WorktreeCreateResult, WorktreeDeleteRequest,
+    WorktreeMoveRequest, WorktreeOpsError,
 };
--- a/crates/tmai-core/src/worktree/ops.rs
+++ b/crates/tmai-core/src/worktree/ops.rs
@@ -10,12 +10,17 @@ use tokio::process::Command;
 use crate::git::is_valid_worktree_name;
 
 use super::types::{
-    WorktreeCreateRequest, WorktreeCreateResult, WorktreeDeleteRequest, WorktreeMoveRequest,
-    WorktreeOpsError,
+    BaseStalenessReport, WorktreeCreateRequest, WorktreeCreateResult, WorktreeDeleteRequest,
+    WorktreeMoveRequest, WorktreeOpsError,
 };
 
 /// Timeout for git worktree commands
 const GIT_TIMEOUT: Duration = Duration::from_secs(10);
+
+/// Timeout for the optional `git fetch origin <base>` performed before
+/// branching. Network operations are slower than local plumbing, so this
+/// gets a longer budget than `GIT_TIMEOUT`.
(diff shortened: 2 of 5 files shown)


trust-delta (commented):
Nicely self-contained change. Validation ordering is correct (`is_safe_git_ref` runs before building `origin/<base>` and the `<base>..origin/<base>` range), the fetch is truly best-effort, tests cover the four relevant states (stale / up-to-date / no origin / disabled), and `dispatch_review` explicitly opts out to avoid redundant work — good instincts. `ActionPerformed` emit looks appropriate here (unusual situation, not routine noise).

A few small observations, none blocking:

1. **Comment is slightly misleading** — `crates/tmai-core/src/worktree/ops.rs:118-125` says "we may be passed something like a SHA — skip the fetch". But `is_safe_git_ref` only rejects empty-string and leading-`-`, so SHAs (`abc123`) pass through and still trigger a `git fetch origin <sha>` + `rev-parse origin/<sha>`. In practice this works because both fail and we fall back to local, so it's harmless — but the comment overstates the guard.

2. **Missing local base is not handled explicitly** — if local `<base>` doesn't exist (e.g. caller passes `main` in a `master`-default repo, or a freshly cloned repo that hasn't materialized the local branch yet), the `rev-list --count <base>..origin/<base>` fails, `behind` becomes `0` via the fallback, and we return `(Some(base), None)`. Subsequent `git worktree add -b <branch> <path> <base>` then errors with a raw git message. Pre-existing behavior, but this PR is the natural place to branch off `origin/<base>` when local doesn't exist at all. Edge case only; fine to defer.

3. **Silent-fallback debuggability** — fetch/verify/rev-list failures are swallowed with no log. `tracing::debug!` on each failure branch would make "why did it fall back" much easier to answer in the field, especially once users report "the stale warning never fires on my repo".

4. **Minor test style** — `crates/tmai-core/src/worktree/ops.rs:1260-1265` binds `_bare_tmp_keep` and `_upstream` with underscore prefixes but then actually uses them (`&_upstream`, `_bare_tmp_keep.path()`). Convention is to drop the `_` when the binding is read; otherwise it reads like "this is intentionally unused".

None of these block the merge.

PR #471, opened 2026-04-16
fix: usage monitor does not auto-start on spawn — manual refresh required (#370)

### LINKED ISSUE #370: fix: usage monitor does not auto-start on spawn — manual refresh required
## Problem

When an agent is spawned, the usage monitor panel does not populate its data automatically. The agent is detected and visible in the WebUI, but the usage widget stays empty (or shows stale data) until the user manually triggers a refresh.

Once a manual refresh is performed (e.g., clicking the refresh button in the usage panel, or hitting \`POST /api/usage/fetch\`), the data loads correctly and continues to update as expected.

## Observed behaviour

1. Spawn an agent (via \`dispatch_issue\`, \`spawn_worktree\`, or the WebUI spawn button)
2. Open the usage panel in the WebUI
3. Widget is blank / stale
4. Click refresh → data appears correctly

## Expected behaviour

The usage fetch should trigger automatically when a new agent is spawned (or at a minimum, on a short periodic interval that doesn't require user intervention).

## Likely areas to investigate

- \`crates/tmai-core/src/usage/fetcher.rs\` — does the fetcher have an auto-start path wired to agent-spawn events?
- \`crates/tmai-core/src/usage/mod.rs\` — service initialisation / scheduling
- \`src/main.rs\` — whether the usage service is spawned at startup and whether it subscribes to \`CoreEvent::AgentAppeared\` or equivalent
- WebUI \`useUsage\` hook (if any) — whether it polls or waits for SSE push

## Workaround

Hit the refresh button in the usage panel after spawning agents.

## Impact

Minor UX friction during dogfooding — users forget the manual refresh step and assume usage tracking is broken.

trust-delta:
## Summary

- Emit `CoreEvent::AgentAppeared` from the poller whenever a new agent is detected (the enum variant existed but was never actually broadcast in production)
- Add `UsageAutoFetchService`: listens for `AgentAppeared` with a 15s debounce to trigger `fetch_usage()`, and runs a 1-minute ticker that honours `usage.auto_refresh_min` so WebUI mode gets periodic refreshes too
- Spawn the service from TUI, WebUI, and tmai-app entry points

Closes #370.

## Test plan

- [x] `cargo test --package tmai --package tmai-core --lib` (1093 tests pass, including 8 new tests for the service gates)
- [x] `cargo clippy --all-targets -- -D warnings`
- [ ] Manual: spawn a new agent and confirm the usage panel populates within ~20s without clicking refresh

## リリースノート

* **新機能**
  * バックグラウンドで自動的に使用量データを更新する機能を追加しました。新しいエージェントが検出されたとき、および設定された間隔で定期的に使用量情報が更新されるようになります。

### DIFF
--- a/crates/tmai-app/src/main.rs
+++ b/crates/tmai-app/src/main.rs
@@ -44,6 +44,10 @@ async fn main() {
             // Auto-fetch usage stats if enabled in settings
             core.start_initial_usage_fetch();
 
+            // Refresh usage when new agents appear + periodic ticker.
+            // Without this the usage widget only updates after a manual refresh (#370).
+            tmai_core::usage::UsageAutoFetchService::spawn(core.clone(), core.subscribe());
+
             // Start event bridge
             let _event_bridge = events::start_event_bridge(core.clone(), app.app_handle().clone());
 
--- a/crates/tmai-core/src/monitor/poller.rs
+++ b/crates/tmai-core/src/monitor/poller.rs
@@ -1540,6 +1540,11 @@ impl Poller {
                     source,
                     initial_status: current_status_name.clone(),
                 });
+                if let Some(ref tx) = self.event_tx {
+                    let _ = tx.send(CoreEvent::AgentAppeared {
+                        target: agent.target.clone(),
+                    });
+                }
                 self.previous_statuses.insert(
                     agent.target.clone(),
                     CommittedAgentState {
--- a/crates/tmai-core/src/usage/mod.rs
+++ b/crates/tmai-core/src/usage/mod.rs
@@ -5,10 +5,12 @@
 
 pub mod fetcher;
 pub mod parser;
+pub mod service;
 pub mod types;
 
 pub use fetcher::{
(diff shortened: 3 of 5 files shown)


trust-delta (commented):
## LGTM (orchestrator-side direct review)

Clean fix. Key points verified:

- `CoreEvent::AgentAppeared` emission in `poller.rs` is in the right place (inside the "new agent detected" branch, gated by `event_tx`).
- `UsageAutoFetchService` correctly skips the first `interval::tick()` so it doesn't race with `start_initial_usage_fetch`.
- 15s spawn debounce is appropriate — running `/usage` spawns a short-lived Claude Code process, so piling up fetches during multi-agent spawn would be wasteful.
- `should_fetch_on_tick` correctly respects `auto_refresh_min == 0` as "manual only", not as "every tick".
- Gate tests (8 new) cover: disabled flag, first-event allow, debounce, fetching-in-progress, tick interval respect.
- Entry points (TUI / WebUI / tmai-app) all spawn the service consistently.

No blockers. Merging via admin squash (branch may be behind after recent merges).

Issue #470, opened 2026-04-16
bug: WebUI Git panel stuck on "Loading branches..." when branch count >= 8


trust-delta:
## Symptom

The Web UI Git panel displays **"Loading branches..."** indefinitely and never renders the branch list. Observed with 8 branches present:

```
334-feat-warn-..., 383-feat-default-..., 439-feat-rewrite-..., main,
pr-467, pr-468, review-pr-467, review-pr-468
```

## Diagnosis

- Backend `list_branches` MCP / HTTP API returns the 8 branches correctly
- Browser DevTools: **no `/api/git/branches` request appears in the network log** even after the panel has been visible for minutes
- No console errors or warnings
- Page was loaded cleanly (no caching suspicion — token-based auth URL)

So either:
1. The initial fetch is never dispatched (useEffect / trigger condition broken at some N)
2. The fetch completes silently but the store doesn't transition out of `loading: true`

## Reproduction

1. Create enough worktrees + review worktrees that total local branches reach roughly 8+
2. Open the WebUI on a fresh tab
3. Expand the Git panel on the right side
4. Panel stays on "Loading branches..." indefinitely

## Likely regression origin

Two recent changes are candidates:

- **#425 / #441** unified worktree/git state onto poller-driven SoT. If initial hydration depends on an SSE event that doesn't fire for this branch list size, the loading flag never flips.
- **#440 / #466** origin-aware notification filtering may have incidentally touched SSE dispatch / frontend event handling in a way that silently drops the payload once the list crosses a threshold.

## Impact

Blocks visibility into branch state from the WebUI — a primary dogfooding surface. Workaround: use `list_branches` MCP tool directly, which is not user-friendly.

## Acceptance criteria

- [ ] Identify whether the issue is no-fetch or no-state-transition (add a breakpoint or a log)
- [ ] Fix the root cause (not a "fire an extra fetch on mount" workaround)
- [ ] Add a component-level test that renders the Git panel with N=10 branches and asserts the list appears
- [ ] Verify in WebUI with the same 8-branch setup

trust-delta:
Moved to https://github.com/trust-delta/tmai-react/issues/1 as part of the repo split. Closing here to consolidate follow-up.

PR #472, opened 2026-04-16
bug: WebUI Git panel stuck on "Loading branches..." when branch count >= 8 (#470)

trust-delta:
## Summary
- **Decouple the panel's loading state from optional fetches.** `BranchGraph#fetchData` awaited a 4-way `Promise.all` across `listBranches`, `gitGraph`, `listPrs`, and `listIssues`. The `gh`-CLI-backed calls have no client-side timeout and can stall indefinitely on auth prompts or rate limits; a single hang there prevented `.finally(() => setLoading(false))` from running, so `loading=true` wedged forever and the panel sat on "Loading branches..." (crates/tmai-app/web/src/components/worktree/BranchGraph.tsx:105-137). Each fetch now writes its own slice of state; `loading` is gated only on `listBranches`.
- **Register the `git_state_changed` SSE listener.** It was missing from `subscribeSSE`'s `namedEvents` array, so every GitMonitor transition added in #423 was silently dropped before reaching `BranchGraph`'s handler. Reconnects the event to the refetch path (crates/tmai-app/web/src/lib/api-http.ts:1190-1210).
- **Two regression tests.** `branch-load-resilience.test.ts` exercises the fetch shape with 10 branches, asserting that a hang on any optional endpoint still resolves the load. `sse-named-events.test.ts` stubs `EventSource` and asserts `git_state_changed` is among the registered listeners.
- **Regenerated web bundle** to match the fresh source; CI drift check verified.

## Test plan
- [x] `pnpm test` — 15 suites / 114 tests pass (2 new suites, 5 new assertions)
- [x] `pnpm lint` / biome — clean, zero warnings
- [x] `pnpm tsc -b` — typechecks clean
- [x] `pnpm build` — rebuilt `src/web/assets/` in-tree
- [x] `cargo check -p tmai -p tmai-core` — clean (unrelated `tmai-app` Tauri crate failure pre-exists on main)
- [ ] Manual verification in WebUI with 8+ branches reproducing the original setup


trust-delta (commented):
## LGTM (orchestrator-side direct review)

Root cause analysis is spot-on — this is a two-bug fix, not a single issue:

1. **`Promise.all` wedging on `gh` CLI stalls** — Correctly decoupled. Each fetch now writes its own state slice; `loading` gates only on `listBranches`. The `.finally` dependency on optional fetches was the actual killer and the fix removes it cleanly.
2. **`git_state_changed` missing from `namedEvents`** — Nice catch. The handler in `BranchGraph` was already wired; the payload was being dropped at the SSE subscription layer (regression from #423).

### Verified

- `BranchGraph.tsx` fetch shape: independent `.then/.catch` per endpoint, only `listBranches` returned as the gating promise. ✓
- `branch-load-resilience.test.ts`: four cases cover the exact regression contract (happy path with N=10, listPrs/listIssues hang, gitGraph hang, listBranches reject). Good regression guard. ✓
- `sse-named-events.test.ts`: Fake `EventSource` sanity-checks `git_state_changed` + `pr_created` + `agents` all registered. Prevents silent dropping from a future `namedEvents` rewrite. ✓
- Web bundle regenerated in-tree, CI drift check confirmed. ✓

### Nits (non-blocking)

- `as Record<string, PrInfo>` / `as IssueInfo[]` casts hint the API layer's return types aren't fully inferred — orthogonal to #470, will naturally improve as #446 (ts-rs + utoipa) rolls out to more endpoints.

No blockers. Merging via admin squash.

Issue #426, opened 2026-04-13
feat: unify Transcript tail via append-event stream


trust-delta:
## Problem

Agent transcript (Claude Code JSONL) has two independent consumers:

- **TranscriptWatcher** runs inside the poller (`crates/tmai-core/src/monitor/poller.rs:232`) tailing the JSONL files
- **`/api/agents/{target}/transcript`** reads the parsed transcript on each request (`crates/tmai-app/web/src/lib/api-http.ts:819`)

The two paths are unsynchronized: under heavy write rate the watcher can lag the file, and the API may return a list that doesn't match what the watcher has digested. UIs see incomplete or out-of-order transcripts during burst writes.

## Goal

Single producer (TranscriptWatcher) → append events → both API and SSE deliver in lockstep.

## Proposed direction

- Watcher emits `TranscriptAppended { agent_id, records }` `CoreEvent` per new chunk
- WebUI subscribes via SSE, builds the transcript incrementally rather than re-fetching
- `/api/agents/{target}/transcript` serves the watcher-maintained in-memory log, not a fresh file read
- Cold start (UI opens after writes already happened) still uses one bulk read; subsequent updates are append-only

## Out of scope

- Long-term storage / search of transcripts (separate concern)

## Priority

Low — current behavior is read-only and missing-records correct themselves on next refresh. File now to round out the unification series.

## Related

- #422 — PR Monitor pattern
- Sibling unification issues for git / worktree / agent state

trust-delta:
Moved to https://github.com/trust-delta/tmai-core/issues/22 as part of the repo split. Closing here to consolidate follow-up.

PR #479, opened 2026-04-17
fix: cap WebUI transcript render to prevent freeze under long sessions

trust-delta:
## Summary

Operators reported the WebUI freezing when selecting a worker pane after the worker had accumulated a few hundred transcript records. Root cause is in the PreviewPanel → TranscriptView pipeline:

1. `TranscriptView.tsx` mapped every record into a `react-markdown + remarkGfm` subtree with no virtualization. At ~500 records the initial mount froze the tab for seconds.
2. `PreviewPanel.tsx::fetchTranscript` ran on a 3s interval and called `setTranscriptRecords(data.records)` with a fresh array on every tick, so even a stable transcript rebuilt the entire subtree 20x/minute.

Observed in dogfooding when two worker agents had transcripts of 144 and 437 records respectively; selecting either panel froze the Chrome App tab until we forced it closed.

## Changes

- **TranscriptView**: render only the last `DEFAULT_VISIBLE_COUNT` (150) records by default; add a "Show N earlier records" toggle that reveals the full history on demand. Operators almost always care about the tail, so the common case is now bounded in DOM size.
- **PreviewPanel.fetchTranscript**: bail out of `setTranscriptRecords` when the polled records array matches the previous one (same length + same tail uuid — sufficient for this append-only transcript). Eliminates the 3s rebuild for idle agents.
- **tests**: added unit tests for the cap logic and the bail-out predicate (alongside the existing `transcript-view.test.ts` helpers). Suite now passes 23 tests.
- Regenerated `src/web/assets/` bundle.

## Why a cap instead of `react-virtual`

- No new dependency; change is ~30 LOC.
- Variable-height markdown items make `react-virtual` configuration non-trivial and would need height measurement to avoid layout thrash.
- #426 (Transcript tail append-event stream) will structurally obsolete the cap by making the render pressure O(delta) instead of O(full history). This PR unblocks dogfooding until that lands.

## Test plan

- [x] `npx biome check` — clean
- [x] `npx tsc -b --noEmit` — clean
- [x] `npx vitest run src/components/agent/__tests__/transcript-view.test.ts` — 23/23 pass
- [x] `npm run build` — bundle regenerated
- [ ] Manual: load WebUI with an agent that has >150 transcript records → panel selects without freeze; "Show earlier" reveals the older records.
- [ ] Manual: leave an agent idle for ~10s → TranscriptView does not flicker / re-render on each 3s poll.

## Related

- #426 feat: unify Transcript tail via append-event stream (root-cause fix)
- Memory: dogfooding-observed freeze during parallel worker dispatch session 2026-04-17

## リリースノート

* **新機能**
  * トランスクリプトビューにデフォルトで最新150件のレコードを表示するキャップ機能を追加しました
  * 「過去のレコードを表示」トグルボタンを実装し、すべてのレコードを表示できるようにしました

* **パフォーマンス改善**
  * トランスクリプト取得時の冗長なReactの状態更新を削減し、レンダリング性能を向上させました

* **スタイル**
  * CSS アセットの更新と最適化を実施しました


trust-delta:
Added a second commit (`9f81233`) to address a parallel freeze path that the transcript cap alone didn't cover.

## Why

Operators reported the Agent panel still freezing after the first commit landed in our local dogfooding build. Investigation showed the transcript cap only covers one of the two heavy render paths inside `PreviewPanel`:

- **Transcript** (structured log) — capped in the first commit.
- **Preview history** (capture-pane ANSI scrollback) — ran the full blob through `shrinkContentToWidth → AnsiUp.ansi_to_html → DOMPurify.sanitize → innerHTML` every time the `history` string changed. For long-running agents the scrollback balloons past 10k lines and freezes the tab on mount.

## What

- `capHistoryLines(raw, maxLines)` — pure helper in `preview-content.ts` that tail-slices history to the last N lines and reports how many were dropped.
- `PreviewPanel` applies `MAX_HISTORY_LINES = 2000` before AnsiUp runs. Live region is untouched.
- Dimmed "… N earlier lines hidden" header surfaces the cap so the truncation isn't silent.
- Unit tests for under/over-cap, empty/non-positive, trailing-newline, and a 25k-line stress case.

Bundle regenerated. Please re-pull and restart tmai to pick up both commits.

PR #487, opened 2026-04-17
fix: narrow branchAgentKey to bare branch set (drop target)

trust-delta:
## Summary

Follow-up to #485 / #486. Hang persisted even after `status` was dropped from the frontend key and statusline counters were removed from the backend fingerprint. `tmai.log` held the smoking gun:

```
INFO agent target migrated via PID reconciliation old_target=hook:0.4 new_target=hook:0.3 pid=2136
```

PID reconciliation flips `target` between hook pane ids while the agent is unchanged. Because `target` was still in `branchAgentKey`, every reconciliation tick flipped the key, invalidated `branchAgentMap` → `nodes` → `layout`, and the LaneGraph repainted.

## Change

`branchAgentKey` now captures **only the set of branches that have an agent attached** — dropping `target` and everything else:

```ts
const branchAgentKey = useMemo(
  () => agents
    .filter((a) => a.git_branch)
    .map((a) => a.git_branch)
    .sort()
    .join("\x02"),
  [agents],
);
```

The layout-affecting question is "does this branch have an agent on it?"; the specific `target` string only matters for the action panel when the user selects that branch, and ActionPanel still reads from live `activeNode` + real `agents` via other channels when it mounts.

## Tradeoff

Node rows may display stale `agentTarget` / `agentStatus` text until the branch set changes (agent spawn/stop on a different branch). Sidebar and ORCH header continue to track target/status live via their own subscriptions.

## Verification

- `npx biome check` / `npx tsc --noEmit` / `pnpm test` (114/114) / `pnpm build` — all green.

## Test plan

- [ ] Rebuild tmai, open tmai project Branch graph with orchestrator active, confirm tab stays responsive during sustained tool-call activity.
- [ ] Confirm sidebar still reflects real-time status / target migrations.
- [ ] If still hanging: investigate non-SSE / non-agents paths (gh subprocess storm, sidebar polling, etc.).


trust-delta:
Wrong direction — the real axis is commit count (LaneGraph renders 200 × lanes DOM nodes on initial mount for the tmai project, vs 7 for small projects). Replacing with a graphLimit default reduction.

PR #492, opened 2026-04-18
bug: compute_branch_parents can produce cyclic parents map → client hang (#491)

### LINKED ISSUE #491: bug: compute_branch_parents can produce cyclic parents map → client hang
## Symptom

The WebUI Branch graph panel could hang the Chrome tab ("Not responding") synchronously on first mount whenever the `parents` map returned by `GET /api/git/branches` contained a cycle (e.g. `A → B → A`). The client-side loop that hit it — `layout.ts:branchDepth` — was guarded in #489, but the upstream source of the cyclic data is still the backend.

## Location

`crates/tmai-core/src/git/mod.rs::compute_branch_parents` (around line 710). It populates a `HashMap<String, String>` mapping each non-default branch to a single parent via two steps per branch:

1. `reflog_created_from` — reads git reflog "Created from X" entries
2. `find_closest_parent_branch` fallback — uses `git for-each-ref --merged=<branch>`

Neither step enforces the invariants a downstream consumer needs:

- `parent != branch` (no self-loop).
- The global mapping forms a DAG (no `A → B → A`, `A → B → C → A`, etc.).

## Why it matters

- Exposes every downstream consumer (WebUI, TUI, any future MCP / API clients) to potential infinite loops unless each one defensively keeps its own visited set. That is brittle — #489 only covered `branchDepth`; there are other walks in `layout.ts` (e.g. `getDepth`, `assignCommitsToBranches`) that *happen* to already guard, but nothing enforces this.
- The contract of the `parents` field is documented-by-accident. Callers reasonably assume it's a tree rooted at `default_branch`.

## How cycles form (hypothesized)

- Worktree branches created from a non-default base followed by branch renames / deletions / recreations can leave reflog entries that round-trip (A's reflog says "created from B", B's reflog says "created from A" for a recreated instance).
- The two detection strategies (`reflog_created_from` vs. `find_closest_parent_branch`) are independent; if they disagree on a pair of branches, the aggregate can cycle.

Real-world trigger on this repo: a `tmai` worktree with 4+ worktree branches + remote-only counterparts. The exact reflog state that caused the cycle wasn't captured — worth doing next time we reproduce.

## Proposed fix

1. After `compute_branch_parents` assembles the raw map, run a validation pass:
   - Drop any entry where `parent == branch` (self-loop).
   - For each entry, walk the chain with a `HashSet<String>` of visited nodes. If the walk revisits a node, break the weakest edge — prefer keeping the reflog-backed link over the `find_closest_parent_branch` fallback, or fall back to `default_branch` if the weakest edge can't be identified.
2. Add a unit test with a fixture that installs a cyclic relationship in the raw map and asserts the post-pass map is acyclic (and rooted at `default_branch`).
3. Emit `tracing::warn!` when a cycle is broken, so we can tell from logs if this happens in the wild.

## Related

- #489 — client-side `visited` guard in `layout.ts:branchDepth` (defensive mitigation, already merged).
- #470 / #472 — earlier Branch graph loading-state resilience.

## Acceptance criteria

- [ ] `compute_branch_parents` never returns a map with self-loops or cycles, regardless of the underlying reflog state.
- [ ] A unit test in `crates/tmai-core/src/git/mod.rs` tests (or a new module) covers the cycle-breaking logic with at least: self-loop, two-node cycle, three-node cycle.
- [ ] A `tracing::warn!` is emitted per broken edge with enough context to debug later.
- [ ] `layout.ts:branchDepth`'s defensive guard stays in place — belt-and-braces is fine.

trust-delta:
## Summary
- Harden `compute_branch_parents` at the source so it can never emit a cyclic `branch -> parent` map. The reflog "Created from" and `for-each-ref --merged` strategies are independent and can disagree (self-loops, two-node cycles, longer cycles) — #489 guarded the WebUI consumer but leaves TUI, MCP, and any future API client vulnerable unless each keeps its own visited set.
- New `break_parent_cycles` pass: drop self-loops and redirect the cycle-closing edge to `default_branch`; every break is logged via `tracing::warn!` for debuggability.
- The `layout.ts:branchDepth` defensive guard from #489 stays in place as belt-and-braces.

## Test plan
- [x] `cargo test -p tmai-core --lib git::` — 46 tests pass, including the live `test_compute_branch_parents_returns_map` and the four new cycle-breaking unit tests (self-loop, two-node cycle, three-node cycle, acyclic-untouched).
- [x] `cargo fmt` clean on the changed file.
- [x] `cargo build -p tmai-core` succeeds.

Closes #491.

### DIFF
--- a/crates/tmai-core/src/git/mod.rs
+++ b/crates/tmai-core/src/git/mod.rs
@@ -797,10 +797,7 @@ fn break_parent_cycles(parents: &mut HashMap<String, String>, default_branch: &s
         let mut visited: HashSet<String> = HashSet::new();
         visited.insert(start.clone());
         let mut cursor = start;
-        loop {
-            let Some(parent) = parents.get(&cursor).cloned() else {
-                break;
-            };
+        while let Some(parent) = parents.get(&cursor).cloned() {
             if visited.contains(&parent) {
                 tracing::warn!(
                     branch = %cursor,
--- a/crates/tmai-core/src/git/mod.rs
+++ b/crates/tmai-core/src/git/mod.rs
@@ -756,7 +756,64 @@ async fn compute_branch_parents(
         }
     }
 
+    break_parent_cycles(&mut parents, default_branch);
+
     parents
+}
+
+/// Sanitize a `branch -> parent` map so it is guaranteed to be acyclic and
+/// rooted at `default_branch`.
+///
+/// Why: the two detection strategies in `compute_branch_parents`
+/// (reflog "Created from", `for-each-ref --merged`) are independent and can
+/// disagree on a pair of branches, producing self-loops (`A -> A`) or cycles
+/// (`A -> B -> A`). Downstream consumers (WebUI `layout.ts`, TUI) walk this
+/// map and previously hung on cycles (see #489). Fix the contract at the
+/// source instead of asking every caller to keep its own visited set.
+///
+/// Strategy: drop self-loops, then for each branch walk the parent chain with
+/// a visited set; if a cycle is detected, redirect the closing edge to
+/// `default_branch` (which has no entry, so the walk terminates). Every
+/// broken edge is logged via `tracing::warn!`.
+fn break_parent_cycles(parents: &mut HashMap<String, String>, default_branch: &str) {
+    // Step 1: collapse self-loops (parent == branch) onto default_branch.
+    let self_loops: Vec<String> = parents
+        .iter()
+        .filter(|(branch, parent)| branch == parent)
+        .map(|(branch, _)| branch.clone())
+        .collect();
+    for branch in &self_loops {
+        tracing::warn!(
+            branch = %branch,
+            "parent map had self-loop; redirecting to default branch"
+        );
+        parents.insert(branch.clone(), default_branch.to_string());
+    }
+
+    // Step 2: for each branch, walk its chain; if it revisits a node, break
+    // the edge that closes the cycle by pointing it at default_branch.
+    let keys: Vec<String> = parents.keys().cloned().collect();
+    for start in keys {
+        let mut visited: HashSet<String> = HashSet::new();
+        visited.insert(start.clone());
+        let mut cursor = start;
+        loop {
+            let Some(parent) = parents.get(&cursor).cloned() else {
+                break;
+            };
+            if visited.contains(&parent) {
+                tracing::warn!(
+                    branch = %cursor,
+                    original_parent = %parent,
+                    "parent map had cycle; redirecting to default branch"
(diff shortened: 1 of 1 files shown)


trust-delta (commented):
## Review: LGTM (cannot self-approve via `gh`; posting as comment)

### Correctness
The two-phase sanitizer produces an acyclic, default-rooted map under all realistic inputs:

- **Self-loop** (`A → A`): caught by Step 1's dedicated pass and redirected to `default_branch`.
- **2-node cycle** (`A → B → A`): Step 2 visits `{start, parent, parent_of_parent}`; when `parent_of_parent` reappears in `visited`, the closing edge at `cursor` is rewritten to `default_branch`.
- **Longer cycles / chain-into-cycle** (e.g. `A → B → C → B`): same mechanism — the revisit always happens at a known node, so rewriting the edge at `cursor` definitively breaks the loop.
- **Missing keys** (parent not in map, e.g. pointing at `default_branch` or an unknown branch): the `while let Some(parent) = parents.get(&cursor)` terminates naturally.
- **Termination**: `default_branch` is never inserted as a key here and is skipped in `compute_branch_parents` (git/mod.rs:719), so any redirected edge terminates on the next hop.

Because the walk mutates entries in place, starting-order doesn't matter — once an edge is redirected to `default_branch`, any subsequent start visiting the same `cursor` follows the already-broken chain.

### Relation to #489
The `branchDepth` visited-set guard in `crates/tmai-app/web/src/components/worktree/graph/layout.ts:93-108` stays untouched. This PR addresses the producer; #489 protects the consumer. Complementary, not duplicative — good defense-in-depth.

### Test coverage
`test_break_parent_cycles_drops_self_loop`, `..breaks_two_node_cycle`, `..breaks_three_node_cycle`, and `..leaves_dag_untouched`, plus the shared `assert_parent_map_is_acyclic` helper, cover the meaningful failure modes. The helper re-verifies the invariant downstream consumers rely on, which is the right regression shape for this class of bug. The DAG test implicitly covers the "parent points at a non-key" path (feat-a → main, main absent from the map).

### Minor observations (non-blocking)
- `test_break_parent_cycles_breaks_two_node_cycle` asserts `redirected >= 1`, but the implementation guarantees exactly one redirect for a pure 2-node cycle. Tightening to `== 1` would lock in that property, though it's fine as-is given iteration-order non-determinism concerns in larger cases.
- Worst case is O(N²) chain walks, bounded by the existing `branches.len() > 100` early-return in `compute_branch_parents`, so no scaling concern.

trust-delta:
@coderabbitai review

trust-delta:
@coderabbitai review

Issue #464, opened 2026-04-15
feat: orchestrator notify policy for contract-layer events


trust-delta:
## Background

See `.claude/research/2026-04-15_contract-layer-architecture-comparison.md` for full architecture context.

The current `OrchestratorNotifier` forwards `ActionPerformed` events to the orchestrator based on a simple origin-filter (silent for agent actions, notify for human actions, etc. — see `feedback_action_notification_noise.md`).

With contract-layer events (G2: `DispatchRejected`, `VendorAvailabilityChanged`, `CapacityChanged`, `ContractViolation`, `DispatchBypassUsed`, `BundleStatusChanged`), naive forwarding would flood the orchestrator with operational noise. But ignoring them entirely loses information the orchestrator needs to make good decisions (e.g., \"CC Max rate-limit cleared — resume queued work\").

A contract-aware notify policy decides *which* contract events reach the orchestrator, *when*, and in *what form* (individual vs aggregated).

## Goal

Define and implement a `NotifyPolicy` that routes contract-layer events to the orchestrator with the right aggregation, de-dup, and escalation rules. Keep the orchestrator's attention focused on state changes it can act on, without drowning it in routine state-flip noise.

### Policy matrix

| Event | Default routing | Rationale |
|---|---|---|
| `DispatchRejected` (orchestrator's own intent) | **notify immediately** | Orchestrator needs to know its dispatch failed to choose alternative (wait, retry, escalate) |
| `DispatchRejected` (human/system origin) | **silent** | Not orchestrator's business |
| `VendorAvailabilityChanged: Available → RateLimited` | **silent (first N), then aggregate** | Gatekeeper will naturally reject; no need to prompt orchestrator pre-emptively |
| `VendorAvailabilityChanged: RateLimited → Available` | **notify** | Resume signal — orchestrator may have pending work to kick off |
| `CapacityChanged` | **silent** | High-frequency, already reflected in gatekeeper decisions |
| `ContractViolation` (orchestrator's action) | **notify** | Audit + chance to reason about it |
| `ContractViolation` (other origins) | **silent unless escalated** | Noise for orchestrator |
| `DispatchBypassUsed` | **silent** | Audit-only, visible in audit log |
| `BundleStatusChanged: → AllGreen` | **notify** | Orchestrator may trigger merge flow |
| `BundleStatusChanged: → PartialFailed / AllFailed` | **notify** | Orchestrator decides rollback / escalation |
[...]

Issue #463, opened 2026-04-15
feat: CoreEvent extensions for contract-layer state


trust-delta:
## Background

See `.claude/research/2026-04-15_contract-layer-architecture-comparison.md` for full architecture context.

`CoreEvent` is today's broadcast bus for agent/team-level changes. With the contract layer (#454/#455/#458/#459/#460) landing, new categories of state change exist that no current event expresses:

- A dispatch was rejected before spawn (the orchestrator never sees it as an agent; no existing event fires)
- A vendor's rate-limit status flipped (auto-resume happened; orchestrator may want to know)
- The global capacity counter moved (UI badge update; may trigger policy decisions)
- A contract violation occurred (for audit / escalation)

Without these events, contract-layer behavior is invisible to subscribers (WebUI, OrchestratorNotifier, MCP clients, audit log).

## Goal

Extend `CoreEvent` with contract-layer variants so every contract-relevant state change is observable through the existing broadcast channel.

### New variants

```rust
pub enum CoreEvent {
    // ... existing variants ...

    /// A dispatch request was rejected by the gatekeeper before spawn
    DispatchRejected {
        intent_summary: DispatchIntentSummary, // redacted intent (no secrets)
        origin: ActionOrigin,
        error: TmaiError,
    },

    /// Vendor availability changed (rate-limit hit or cleared)
    VendorAvailabilityChanged {
        vendor: String,
        account: Option<String>,
        old: VendorAvailabilityState,
        new: VendorAvailabilityState,
        detected_via: DetectionSource,
    },

    /// Global capacity state changed (slot consumed or freed)
    CapacityChanged {
        current: usize,
        limit: usize,
        delta: i32,           // +1 on spawn, -1 on agent terminal
        cause: CapacityCauseSummary,
    },

    /// A contract-layer violation occurred (audit event)
    ContractViolation {
        origin: ActionOrigin,
        code: ErrorCode,
        context: serde_json::Value,
    },

    /// A bypass flag was used for a dispatch (audit event)
    DispatchBypassUsed {
        origin: ActionOrigin,
        bypassed: Vec<String>,  // names of bypassed steps
        reason: String,
    },

    /// Bundle status changed (rolled up from sub-dispatches)
    BundleStatusChanged {
        bundle_id: BundleId,
        old: BundleStatus,
        new: BundleStatus,
    },
}
```

### Payload design principles

- **Redaction**: event payloads never include secrets, tokens, or raw prompts. Use summaries and references.
[...]

Issue #460, opened 2026-04-15
feat: role x model presets for agent spawn


trust-delta:
## Background

Anthropic's official Claude Code Agent Teams has a well-documented community pain point: **all teammates run the same model with the same tool set**. Role-based model selection (lead runs Opus, implementers run Sonnet, reviewers run Codex) is a frequently requested feature that hasn't landed.

tmai, being multi-vendor from day one, can natively mix Claude Opus, Claude Sonnet, Codex (GPT), Gemini, and future CLIs — but today the vendor/model combination is specified ad-hoc per spawn (env vars, CLI flags, manual prompts). There's no reusable, declarative way to say \"for this project, the orchestrator is Opus, implementers are Sonnet, reviewers are Codex.\"

Formalizing this as **role presets** is arguably the single most concrete differentiator vs. Anthropic's Agent Teams and single-vendor tools.

## Goal

Declarative `[[role_preset]]` configuration that binds a role name to a vendor, model, system-prompt template, and MCP tool allowlist. Spawn APIs accept a role name and compose the full worker configuration from the preset.

### Settings

```toml
[[role_preset]]
name = \"orchestrator\"
vendor = \"claude\"
model = \"opus-4-6\"
system_prompt_template = \"You are the orchestrator agent. Coordinate workers via MCP tmai tools. {{workflow_rules}}\"
mcp_allowlist = [\"list_agents\", \"spawn_agent\", \"dispatch_issue\", \"dispatch_review\", \"approve\", \"send_prompt\", \"get_transcript\", \"list_prs\", \"list_issues\"]

[[role_preset]]
name = \"implementer\"
vendor = \"claude\"
model = \"sonnet-4-6\"
system_prompt_template = \"You are an implementer. Issue: {{issue_ref}}. {{project_rules}}\"
mcp_allowlist = []  # worker has no MCP access by default

[[role_preset]]
name = \"reviewer\"
vendor = \"codex\"
model = \"gpt-5\"
system_prompt_template = \"You are a code reviewer. PR: {{pr_ref}}. Focus on {{review_focus}}.\"
mcp_allowlist = [\"get_pr_comments\"]

[[role_preset]]
name = \"researcher\"
vendor = \"gemini\"
model = \"gemini-2.5-pro\"
system_prompt_template = \"You are a research agent. Question: {{question}}.\"
mcp_allowlist = []
```

### Template variables

[...]

Issue #459, opened 2026-04-15
feat: cross-repository dispatch (DispatchBundle)


trust-delta:
## Background

#457 (cross-workspace integration) gives tmai a **view** across multiple projects. The next step is **action**: dispatching a single logical task that spans multiple repositories simultaneously.

Concrete triggers:
- Brand-unified release: one change affecting `acme-web`, `acme-api`, `acme-mobile`, and `acme-docs` — four repos, four PRs, coordinated merge
- Shared library upgrade: bump `@acme/ui` across all consumer repos with synchronized smoke tests
- Security patch: apply the same fix pattern to 5 services

Today the human operator manually dispatches 4+ separate `dispatch_issue` calls, tracks them by hand, and correlates results mentally. This is exactly the human-as-hub problem the orchestration moat is meant to solve.

## Goal

Introduce a first-class **DispatchBundle** primitive that launches related sub-dispatches across multiple repositories with shared context, and keeps them relateable for the bundle's lifetime.

### Data model

```rust
struct DispatchBundle {
    id: BundleId,
    title: String,
    context_prompt: String,        // shared preamble injected into every sub-dispatch
    sub_dispatches: Vec<SubDispatch>,
    status: BundleStatus,          // Pending / InProgress / AllGreen / PartialFailed / AllFailed
    created_by: ActionOrigin,
    created_at: DateTime,
}

struct SubDispatch {
    project_path: String,
    agent_id: AgentId,
    base_branch: String,
    prompt: String,                 // per-repo prompt (appended to bundle context_prompt)
    pr_number: Option<u64>,         // filled as work progresses
    status: SubDispatchStatus,
}
```

### MCP tool

```
dispatch_bundle({
    title: \"Acme Q2 release\",
    context_prompt: \"Coordinated release — version 2.0, see RFC-42 ...\",
    sub_dispatches: [
      { project_path: \"/.../acme-web\",    base_branch: \"main\", prompt: \"...\" },
      { project_path: \"/.../acme-api\",    base_branch: \"main\", prompt: \"...\" },
      { project_path: \"/.../acme-mobile\", base_branch: \"main\", prompt: \"...\" },
    ]
})
```

Returns the `BundleId`. Each sub-dispatch is a normal worker with a normal worktree — the bundle is just a correlation layer.

### Orchestrator integration

- `list_bundles`, `get_bundle` MCP tools
- `BundleStatus` rolls up from sub-dispatch states
[...]

PR #493, opened 2026-04-18
feat: CoreEvent extensions for contract-layer state (#463)

trust-delta:
## Summary
- Adds six new `CoreEvent` variants for contract-layer observability: `DispatchRejected`, `VendorAvailabilityChanged`, `CapacityChanged`, `ContractViolation`, `DispatchBypassUsed`, `BundleStatusChanged`.
- Introduces helper types `BundleId`, `DispatchIntentSummary`, `VendorAvailabilityState`, `CapacityCauseSummary`, `BundleStatus` — all `ts-export`-ready and re-exported from `tmai_core::api`.
- Extends the SSE forwarder (`src/web/events.rs`) and the Tauri bridge (`crates/tmai-app/src/events.rs`) to cover the new variants so the workspace still compiles exhaustively.

## Scope
Strictly additive, surgical PR as requested:
- **No emit sites** wired — those land with the features in #454/#455/#459/#460 and the `AutoActionExecutor` follow-up (#364/#464).
- **No notifier policy** (G3) or MCP event-subscription tool.
- Payload design follows the issue (redacted `DispatchIntentSummary`, self-contained `CapacityCauseSummary`, bundle-level rollup only).
- Optional fields use `#[serde(default)]` for forward compatibility, per the issue guidance.

## Design notes
- JSON shape is pinned by eight new unit tests (`crates/tmai-core/src/api/events.rs`), which assert the `type` discriminator, snake_case variant naming for `VendorAvailabilityState` / `CapacityCauseSummary` / `BundleStatus`, omitted fields when `None`, and `DispatchIntentSummary` round-trip with defaults.
- `DetectionSource` gains a `ts_rs::TS` derive under the `ts-export` feature so `CoreEvent::VendorAvailabilityChanged { detected_via: DetectionSource }` is exportable alongside the rest of the enum. No behaviour change outside `ts-export`.
- `ContractViolation { context }` uses the same `skip_serializing_if = serde_json::Value::is_null` + `ts(type = "unknown")` pattern as `TmaiError.context`.

## Test plan
- [x] `cargo fmt --all --check`
- [x] `cargo clippy -p tmai-core --lib -- -D warnings` (lib is clean; the 25 pre-existing test-only clippy warnings on base branch are unrelated)
- [x] `cargo build --workspace`
- [x] `cargo build --workspace --features tmai-core/ts-export`
- [x] `cargo test -p tmai-core --lib` — 1122 passed (18 in `api::events::tests` including 8 new contract-layer shape tests)

## Related
- Issue #463
- Contract-layer siblings: #454, #455, #458, #459, #460
- Consumer: `AutoActionExecutor` (#364), notifier integration (#464)

## 改善

* イベント報告機能を拡張し、ディスパッチ拒否、ベンダー可用性の変更、リソースキャパシティの更新、契約違反、バンドルステータスの変更に関する詳細なシステムイベント情報が取得可能になりました。


trust-delta (commented):
## Review

Scope of this PR matches #463: 6 new `CoreEvent` variants + 5 support structs (`BundleId`, `BundleStatus`, `CapacityCauseSummary`, `DispatchIntentSummary`, `VendorAvailabilityState`), no emission sites yet. Both Rust consumers (TUI `src/ui/app.rs`, web SSE `src/web/events.rs`) and the Tauri bridge (`crates/tmai-app/src/events.rs`) are updated to handle the new arms. The "events-only, emitters land with each feature" stance is documented in-line at `src/web/events.rs:380` — clearly within scope.

The serde shapes look right and the JSON-shape tests in `crates/tmai-core/src/api/events.rs` pin the wire contract well — those tests double as the SSE/TS contract guard, which is the right call for a stability-sensitive surface.

### Issues / questions

**1. TS optionality drift on `Option<T>` fields with `skip_serializing_if`** (medium)

Pattern repeats across the new types. ts-rs renders `Option<T>` as `T | null` (a *required* field that may be null), but `#[serde(default, skip_serializing_if = "Option::is_none")]` makes the wire **omit** the field when `None`. The test at `events.rs:271` even asserts `v["intent_summary"].get("pr_number").is_none()`, so this is by design. Net effect: TS consumers reading e.g. `summary.pr_number` get `undefined`, but the generated type says `bigint | null`.

Affected:
- `DispatchIntentSummary.bundle_id`, `prompt_hash`, `issue_number`, `pr_number`
- `VendorAvailabilityChanged.account`
- `VendorAvailabilityState::RateLimited.resume_at`
- `ContractViolation.context` (typed `unknown`, but disappears entirely when `Value::Null`)
- Inherited from #458 and now first-shipped here: `TmaiError.retry_hint`, `trace_id`, `context`

Two clean fixes — pick one and apply consistently:
- Add `#[cfg_attr(feature = "ts-export", ts(optional))]` so generated TS becomes `pr_number?: bigint | null`. Wire stays the same.
- Or drop `skip_serializing_if` so the field is always serialized as `null`, matching the existing TS shape. Slightly bigger payload but no consumer surprise.

The first option keeps payloads compact and is more honest about the contract. Worth fixing in this PR while the variants have zero emitters — once SSE traffic starts flowing, fixing it later forces a coordinated TS consumer update.

**2. `CapacityChanged` / `BundleStatusChanged` missing `ActionOrigin`** (question)

Per the policy+orchestrator architecture (`.claude/decisions/2026-04-09-policy-orchestrator-rearchitecture.md`), Human/Agent/System origins are tracked through side-effect APIs and surfaced via `CoreEvent::ActionPerformed { origin, ... }`. `DispatchRejected`, `ContractViolation`, and `DispatchBypassUsed` correctly carry `origin`. `CapacityChanged` and `BundleStatusChanged` do not.

Argument for omission: both are downstream *consequences* of an action whose `origin` already shipped via `ActionPerformed`, so attribution is derivable via correlation. Argument against: consumers (e.g. WebUI badges) would otherwise have to keep a sliding window of recent `ActionPerformed`s and correlate by target/timestamp.

Was this deliberate, or worth attaching `origin: ActionOrigin` to `CapacityChanged` (and possibly `BundleStatusChanged`) for symmetry with the rest of the contract-layer set?

**3. `DetectionSource` reuse for vendor availability** (minor)

`DetectionSource` was modeled for *agent state* detection (`HttpHook`, `IpcSocket`, `WebSocket`, `CapturePane`). On `VendorAvailabilityChanged.detected_via`, none of these map cleanly to vendor-side signals (polling, vendor webhook, explicit user signal). Likely needs new variants once emission lands. Not a blocker — flagging so the future emitter PR doesn't quietly stuff `HttpHook` into a slot it doesn't fit.

### Smaller notes

- `CapacityCauseSummary::AgentTerminal` reads as a state name; `AgentTerminated` would parallel `AgentSpawned` better. Wire shape is `agent_terminal` either way — easy to fix now, painful after consumers ship.
- `account: Option<String>` and `context: serde_json::Value` carry `#[serde(default, ...)]`, but `CoreEvent` is `Serialize`-only — the `default` is unused. Harmless, but if you ever flip on `Deserialize` for round-trip tests, you'll want to keep them. Fine as is.
- `#[non_exhaustive]` is not on `CoreEvent`. All consumers I checked use wildcard arms, so no compile breakage today, but worth a follow-up to lock down the contract.

### Verdict

Variant payloads, serde tags, and TS generation are in good shape; the test suite pins the wire format usefully. The only thing I'd want fixed before this merges is the `Option<T>` / TS optionality drift (item 1) — it's a small mechanical change and the right time to do it is *before* anyone subscribes to these events.

trust-delta:
Addressed both must-fixes in `3fd0d82`.

### Fixed
1. **TS optionality drift** — added `#[cfg_attr(feature = "ts-export", ts(optional))]` on every `Option<T>` field that combined `skip_serializing_if = "Option::is_none"`. Regenerated TS: fields now render as `field?: T` instead of `field: T | null`, matching the wire shape (field omitted on `None`).
   - `DispatchIntentSummary.{bundle_id, prompt_hash, issue_number, pr_number}`
   - `CoreEvent::VendorAvailabilityChanged.account`
   - `VendorAvailabilityState::RateLimited.resume_at`
   - `TmaiError.{retry_hint, trace_id}` (the `context: serde_json::Value` fields keep `ts(type = "unknown")` — `ts(optional)` errors on non-`Option<T>`; `unknown` already accepts `undefined`. Documented inline.)
2. **Rename** `CapacityCauseSummary::AgentTerminal` → `AgentTerminated`. Wire form is now `agent_terminated`, paralleling `agent_spawned`.

Wire-format tests (`api::events::tests`, 18 total) still pass unchanged — confirms only the TS declaration moved. `cargo fmt` + `cargo clippy -p tmai-core --lib -- -D warnings` + `cargo test -p tmai-core --lib` all green (1122 passed). `scripts/generate-types.sh` output committed.

### Deferred to follow-ups (not addressed here, per instructions)
- CoderRabbit's `DetectionSource` PascalCase vs snake_case concern — downstream compat.
- Missing `ActionOrigin` on `CapacityChanged` / `BundleStatusChanged` — **deliberate**: those events are deltas attributable by correlation with the nearby `ActionPerformed` / `DispatchRejected` emissions rather than duplicating origin on every capacity tick. Happy to file a follow-up if reviewers want origin inlined.
- `DetectionSource` fit for `VendorAvailabilityChanged.detected_via` — revisit when the emission site lands (it may need a dedicated enum).
- `#[non_exhaustive]` on `CoreEvent` — follow-up.

PR #638, opened 2026-05-10
fix(react): rewrite arrow keys to SS3 so CC's ghost-text accept fires

trust-delta:
## Summary

Pressing → on an Ink-based TUI like CC inside tmai's WebUI was a no-op where the same key in a normal terminal accepts the ghost-text autosuggestion in full. The fix rewrites CSI arrow / Home / End sequences (`\x1b[C`, …) into their SS3 (`\x1bOC`, …) equivalents on egress to the keys WebSocket.

### Why CC misses CSI

CC's key handler table is wired to SS3 sequences because outside of tmai it sits behind tmux, which serves SS3 arrow sequences regardless of whether the inner program ever flipped DECCKM. tmai's PTY-server hands xterm.js the agent's raw stream, and xterm.js (correctly) ships CSI sequences when the agent never enabled application cursor mode. Every byte is technically valid; the binding just never matches.

### Why pre-flipping DECCKM doesn't work

Calling `term.write("\x1b[?1h\x1b[?66h")` on mount switches xterm into application cursor / keypad mode briefly, but CC reliably emits its own DECCKM disable shortly after attaching, so the modes flip back to normal before the user can type. Verified by hooking `term.write` and watching the modes object via React fiber walk.

### What this PR does

- `useTerminal.ts` lifts a single `remapArrowsToSs3` helper to module scope and uses it from both the mount-time `term.onData` and the `setAttachable` re-attach (select-mode → input-mode toggle). Built via `String.fromCharCode(0x1b)` so biome's `noControlCharactersInRegex` is satisfied.
- `keys.ts` (PreviewPanel keys path, which doesn't go through xterm) gets the same SS3 byte sequences directly so both panels behave identically.

PageUp / PageDown / Delete have no SS3 form and stay CSI.

## Test plan

- [x] `pnpm test` — 323 pass / 2 skipped, including new "rewrites CSI arrow sequences to SS3 on egress" test
- [x] `pnpm lint` — clean
- [x] `pnpm exec tsc -b` — clean
- [x] Browser verify with `mcp__chrome-devtools` on a freshly-spawned CC agent:
  - Before fix: → ships `1b 5b 43` (CSI C); ghost-text `Try "refactor poller.rs"` shows on a separate line and a single `T` glyph appears in the input row (CC's preview-cursor advance, not an accept)
  - After fix: → ships `1b 4f 43` (SS3 C); the entire ghost-text lands in the input row and the suggestion line collapses
- [ ] Dogfood: spawn fresh CC inside tmai, accept ghost suggestion via →

### DIFF
--- a/clients/react/src/hooks/__tests__/useTerminal.test.ts
+++ b/clients/react/src/hooks/__tests__/useTerminal.test.ts
@@ -24,6 +24,7 @@ interface FakeTerminalInstance {
   cols: number;
   loadAddon: ReturnType<typeof vi.fn>;
   open: ReturnType<typeof vi.fn>;
+  write: ReturnType<typeof vi.fn>;
   onData: ReturnType<typeof vi.fn>;
   onBinary: ReturnType<typeof vi.fn>;
   onResize: ReturnType<typeof vi.fn>;
@@ -50,6 +51,7 @@ vi.mock("@xterm/xterm", () => {
       cols: 120,
       loadAddon: vi.fn(),
       open: vi.fn(),
+      write: vi.fn(),
       onData: vi.fn(() => makeDisposable()),
       onBinary: vi.fn(() => makeDisposable()),
       onResize: vi.fn((cb: ResizeCallback) => {
@@ -88,6 +90,7 @@ globalThis.ResizeObserver = function ResizeObserver() {
 } as unknown as typeof ResizeObserver;
 
 import { api } from "@/lib/api";
+import { useAgentTerminalStream } from "../useAgentTerminalStream";
 import { useTerminal } from "../useTerminal";
 
 function makeContainerRef(): React.RefObject<HTMLDivElement> {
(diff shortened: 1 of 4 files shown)


trust-delta:
Closing — empirical re-test showed this fix does not solve the user-visible problem.

User confirmed that pressing → on ghost text no longer accepts in **either** tmai's WebUI or a bare `tmux + CC` reference setup ("何も起きませんでした。前は ghost が入ったのですが"). This means CC v2.1.138 itself stopped binding ArrowRight to "accept ghost-text suggestion" (regardless of CSI vs SS3 on the wire) — it's a CC behavior change, not a tmai-side wire-shape mismatch.

Without a real binding to chase, swapping CSI arrows for SS3 here just changes byte-level output that was already working for cursor moves and gives nothing back. Reverting locally; if CC restores the binding (or a different shortcut like Tab / Ctrl+E supersedes it) we can revisit.

PR #699, opened 2026-05-16
feat(webui): codemod + regression guard + settings/ pilot (PR B)

trust-delta:
**PR B of the semantic-token migration** (follows #698). Builds the reusable migration machinery and migrates the first area as the pilot.

## Machinery

- **`scripts/theme-codemod.mjs`** — the **canonical** hardcoded-palette → semantic-token mapping. Every per-area PR (C–H) runs this same dictionary so the swap is consistent repo-wide. Regex codemod with boundary look-around: variant prefixes (`hover:`/`focus:`) and opacity suffixes (`/40`, `/[0.02]`) sit outside each match and survive automatically. Lives outside `src/` (not tsc/biome-scoped). Usage: `node scripts/theme-codemod.mjs [--write] <dir…>`.
- **`src/lib/__tests__/no-raw-palette.test.ts`** — regression guard. Scans an allowlist of migrated areas (`MIGRATED`, currently `["components/settings"]`) and fails if any raw Tailwind palette utility reappears there. The allowlist grows one entry per area PR; the final PR scans `components` wholesale.

## Pilot: `components/settings/`

- 19 files, **312 substitutions, zero residual** raw palette (guard green). Toggle ternaries / component logic untouched — className string content only.
- **`zinc` retuned to be faithful**: its token values are now the *actual* Tailwind palette colours the components hardcoded (zinc-200 `#e4e4e7`, zinc-400 `#a1a1aa`, zinc-600 `#52525b`, cyan-400 `#22d3ee`, red-400 `#f87171`, amber/emerald/blue-500, white/N overlays). The old shadcn `@theme` OKLch values were **never consumed** by components — that was the original "theme doesn't apply" bug; the `theme.ts` comment now states this plainly.
- Under **tokyonight/light**, `settings/` now actually re-skins (the intended effect). Under **zinc** the swap is visually near-identical.

### Honest deltas (intentional, documented in theme.ts)

The canonical mapping collapses shade ranges, so a few off-shade usages shift sub-perceptually under `zinc`: `text-zinc-500` → `muted-foreground` (#a1a1aa, slightly lighter than #71717a) and `bg-white/[0.02]` → `surface` (0.05). Dominant shades are exact. This is "faithful near-identical", not literal pixel-identity — the agreed consequence of consolidating to semantic tokens.

## Gate

`pnpm build && pnpm lint && pnpm test` — green (64 files, **493 passed**, 2 pre-existing skips, incl. the new guard).

**Browser-verify deferred**: the `vite` dev server is a documented WSL-crash risk here and the backend isn't running, so auto-launching it is unsafe/low-value. Structural correctness is covered by build + the zero-residual guard + the static colour analysis above; interactive tokyonight-vs-zinc screenshotting is recommended at review or on request.

Next: PR C (worktree/ — the largest area) using the same codemod + guard entry.

## リリースノート

* **新機能**
  * スタイル自動変換ツール：アプリケーションのカラーパレット全体を統一的なトークンシステムに自動変換
  * 変換後の整合性を検証するテストを追加

* **スタイル**
  * 設定画面全体のカラースキームと視覚デザインを統一的なシステムに更新
  * テーマトークンの色値を調整し、UI 全体の外観統一性を強化

[Review Change Stack](https://app.coderabbit.ai/change-stack/trust-delta/tmai/pull/699?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/clients/react/scripts/theme-codemod.mjs
+++ b/clients/react/scripts/theme-codemod.mjs
@@ -1,0 +1,130 @@
+#!/usr/bin/env node
+// Theme-token codemod for the WebUI semantic-token migration.
+//
+// Rewrites the hardcoded Tailwind palette utility classes
+// (`text-zinc-300`, `bg-white/10`, `text-cyan-400`, …) to the semantic
+// theme tokens introduced in PR A (`text-foreground`, `bg-surface-strong`,
+// `text-primary`, …) so the component tree actually re-skins with the
+// active theme.
+//
+// This is the CANONICAL mapping for the migration — every per-area PR runs
+// this same dictionary so the swap is consistent across the codebase. The
+// `zinc` theme's token values are tuned (in lib/theme.ts) so that this
+// substitution is near-identical under the legacy `zinc` theme; under
+// `tokyonight` / `light` it is the intended re-skin.
+//
+// Why a regex codemod and not ts-morph: classes live in string / template
+// literals and `clsx`-style ternaries; a className-token regex with
+// boundary look-around preserves variant prefixes (`hover:`, `focus:`,
+// `md:`) and opacity suffixes (`/40`) automatically because they sit
+// OUTSIDE each match. The ~30% of cases this can't safely decide
+// (gradients, ambiguous purples) are intentionally left for manual fixup
+// and caught by the regression guard.
+//
+// Usage:
+//   node scripts/theme-codemod.mjs <dir...>            # dry-run (counts)
+//   node scripts/theme-codemod.mjs --write <dir...>    # apply in place
+//
+// Skips `__tests__` and only touches *.tsx / *.ts.
+
+import { readdirSync, readFileSync, statSync, writeFileSync } from "node:fs";
+import { join } from "node:path";
+
+// Each rule: [RegExp, replacement]. The RegExp matches a whole colour
+// utility token; `(?<![\w-])` / `(?![\w/-])` boundaries keep us off
+// partial matches and leave any `variant:` prefix and `/alpha` suffix
+// untouched (they're outside the match), so `hover:text-zinc-300` →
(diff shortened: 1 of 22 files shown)


trust-delta:
@coderabbitai Both findings addressed in 35b2b15:

**1. HandoffThresholdSection identical-branch ternary (nitpick) — fixed, but not by collapsing.**
The codemod's canonical mapping sends both `zinc-400` and `zinc-500` → `muted-foreground`, so the original `disabled ? "text-zinc-500" : "text-zinc-400"` (a deliberate *dimmer-when-disabled* cue) collapsed to a no-op. Collapsing to a single class would keep that UX cue dropped, so instead the distinction is restored through the token tier: `disabled ? "text-subtle-foreground" : "text-muted-foreground"` (subtle = the dimmer tier; preserves the affordance under every theme). A permanent guard test (`no-raw-palette.test.ts`) now fails on any identical-branch class ternary in a migrated area, so the remaining area PRs (C–H) catch this codemod-collapse artifact automatically rather than relying on review. Swept `components/settings/` — this was the only occurrence.

**2. `placeholder-subtle-foreground` unsupported under Tailwind v4 (actionable/major) — false positive, verified empirically.**
The built CSS contains:
```
.placeholder-subtle-foreground::placeholder{color:var(--color-subtle-foreground)}
```
Tailwind v4.3 *does* emit `placeholder-*` color utilities for `@theme` color tokens (the v4 change was to `placeholder-opacity-*`, not `placeholder-{color}`; the pre-existing code already used `placeholder-zinc-600` successfully under v4). No change needed — the migrated inputs render placeholder colour correctly. (Noting the analysis chain in the original comment was inconclusive: the `rg --type=tsx` invocations errored out, so the finding fell back to the v3-vs-v4 web result.)

Gate green: build, lint, vitest 494 passed / 2 pre-existing skips.

